PR #118, opened 2026-04-12
feat(api): add idempotencyKey to POST /bookings to prevent duplicate bookings

### LINKED ISSUE #84: api: add idempotencyKey to POST /bookings to prevent duplicate bookings on retry
## Problem

`POST /bookings` has no idempotency mechanism. If a client's request succeeds but the response is lost (network timeout, worker restart, mobile connection drop), the client retries and creates a **second identical booking**.

The existing `bookings_no_overlap` exclusion constraint prevents *overlapping* bookings for the same vehicle, but it does NOT prevent the same renter from booking the **exact same time range twice** — the constraint sees two bookings with identical `[startAt, effectiveEndAt)` ranges, and `&&` (overlap) on two identical ranges is true, so it rejects. But if there's even a 1-second difference in the retry (client recalculates `startAt` from "now"), the constraint passes and a duplicate is created.

More importantly, the overlap constraint only fires per-vehicle. Two bookings for different vehicles from the same click are not caught at all.

### Real-world scenario

1. Renter taps "Book Now" on a 3-day rental (¥36,000).
2. Server creates booking, returns 201.
3. Response is lost (CF Worker cold-start timeout, mobile tunnel drop).
4. Client retries the same request.
5. Server creates a second booking for the same renter, same dates, same vehicle.
6. Renter is now double-booked and owes ¥72,000.
7. Owner sees two bookings for the same car and has to manually cancel one.

At 200 users this is a support ticket. At 2000 users this is a weekly fire.

## Fix

The schema design doc (`docs/plans/2026-04-07-schema-api-design.md`) already specifies an `idempotencyKey` column on bookings. Implementation:

1. **Schema**: Add `idempotencyKey: text('idempotencyKey')` to the bookings table with a partial unique index: `CREATE UNIQUE INDEX bookings_idempotency ON bookings(idempotencyKey) WHERE idempotencyKey IS NOT NULL`.
2. **Validator**: Add `idempotencyKey: z.string().uuid().optional()` to `createBookingSchema`.
3. **Route**: In `POST /bookings`, if `idempotencyKey` is present:
   - Check if a booking with that key already exists.
   - If yes, return the existing booking (200, not 201) — the response the client missed.
   - If no, create as normal and store the key.
4. **Web client**: Generate a UUIDv4 per booking attempt in the server action. Store it in a ref so retries send the same key.

### Named pattern

This is the **Non-Idempotent POST** anti-pattern. Every state-changing endpoint that a client might retry needs an idempotency mechanism. The three common approaches:

| Approach | Pros | Cons |
|----------|------|------|
| Client-generated idempotency key (UUID) | Simple, stateless server | Client must remember the key across retries |
| Server-generated + return on retry | Server controls key format | Requires a lookup before every create |
| Database unique constraint on natural key | Zero app code | Only works if a natural key exists (it doesn't here — same renter can book same car for same dates intentionally after cancelling) |

We use approach 1 (client UUID) because it's the simplest and the web server action already has a natural "per-click" lifecycle where a ref or closure can hold the key.

## Definition of done

- [ ] `idempotencyKey` column added to bookings schema + migration generated
- [ ] Partial unique index on `idempotencyKey WHERE NOT NULL`
- [ ] `POST /bookings` returns existing booking on duplicate key (200, not 201)
- [ ] Web server action generates and sends a UUID per booking attempt
- [ ] Integration test: two POSTs with same idempotencyKey → same booking ID returned
- [ ] Integration test: two POSTs with different keys → two separate bookings

## References

- Stripe's idempotency key design: the gold standard for this pattern
- `docs/plans/2026-04-07-schema-api-design.md` — already specifies the column

Surfaced during architect review of #80.

jackli921:
## Summary

- **Correctness fix (HIGH):** Network retries on booking creation could produce duplicate bookings. The server now accepts an optional `idempotencyKey` (UUID) and returns the original booking (200) if it sees the same key twice.
- Schema migration adds nullable `idempotencyKey` column + partial unique index.
- Web server action generates a UUID per booking intent automatically.
- Applied cleanly on top of the service layer refactor (#99).

## What changed

| Area | Change |
|------|--------|
| `packages/shared/src/db/schema.ts` | Add `idempotencyKey: text()` to bookings |
| `drizzle/0010_*.sql` + `0011_*.sql` | Column + partial unique index migrations |
| `packages/shared/src/validators/booking.ts` | Add `idempotencyKey: z.string().uuid().optional()` |
| `packages/api/src/stores.ts` | Add field to `Booking` interface |
| `packages/api/src/repositories/*` | Add `findByIdempotencyKey()` to interface + both impls |
| `packages/api/src/services/booking.ts` | Idempotency check before vehicle lookup; returns existing on dup |
| `packages/api/src/routes/bookings.ts` | Pass key through to service; use result status |
| `packages/web/src/lib/bookings.ts` | Generate UUID per booking intent |
| Tests | 3 new integration tests + select-columns update |

## Test plan

- [x] Duplicate idempotencyKey returns same booking ID (200, not 201)
- [x] Different keys create separate bookings (both 201)
- [x] Omitted key still works (backward compat, 201)
- [x] 472 total tests green (119 shared + 137 api + 216 web)
- [x] Pre-commit hooks pass (biome, file-size, module-boundaries, typecheck)
- [x] db:verify clean (12 migrations)

Closes #84

### DIFF
--- a/drizzle/0010_add-idempotency-key.sql
+++ b/drizzle/0010_add-idempotency-key.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "bookings" ADD COLUMN "idempotencyKey" text;
--- a/drizzle/0011_idempotency-unique-index.sql
+++ b/drizzle/0011_idempotency-unique-index.sql
@@ -1,0 +1,3 @@
+CREATE UNIQUE INDEX "bookings_idempotency_key"
+  ON "bookings" ("idempotencyKey")
+  WHERE "idempotencyKey" IS NOT NULL;
--- a/drizzle/meta/0010_snapshot.json
+++ b/drizzle/meta/0010_snapshot.json
@@ -1,0 +1,726 @@
+{
+  "id": "c4e28842-d649-40bd-b60a-57efc2cdf7aa",
+  "prevId": "407c3793-8981-40fd-abc9-b7c1628b26c8",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.accounts": {
+      "name": "accounts",
+      "schema": "",
+      "columns": {
+        "userId": {
+          "name": "userId",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "type": {
+          "name": "type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "provider": {
+          "name": "provider",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "providerAccountId": {
+          "name": "providerAccountId",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "refresh_token": {
+          "name": "refresh_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "access_token": {
+          "name": "access_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "expires_at": {
+          "name": "expires_at",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "token_type": {
+          "name": "token_type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "scope": {
+          "name": "scope",
+          "type": "text",
+          "primaryKey": false,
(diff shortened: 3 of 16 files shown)


jackli921:
Closing — migration number collision with #120 (FK indexes). Will re-apply on a fresh branch.

PR #168, opened 2026-04-12
fix(validators): add cross-field validation to updateVehicleSchema

### LINKED ISSUE #141: fix(validators): updateVehicleSchema.partial() loses cross-field validation
## Problem
`updateVehicleSchema = vehicleObjectSchema.partial()` drops both `superRefine` checks from `createVehicleSchema`. A PATCH setting both rates to null passes validation. `minRentalHours > maxRentalHours` has no DB constraint and would silently persist.

**File:** `packages/shared/src/validators/vehicle.ts:56`

## Fix
Add `superRefine` to update schema that applies cross-field checks when both fields are present.

## Severity: MEDIUM
From architecture audit `docs/2026-04-12-architecture-audit.md`

jackli921:
## Summary
- `updateVehicleSchema = vehicleObjectSchema.partial()` dropped both `superRefine` checks from `createVehicleSchema`
- A PATCH setting both rates to null passed validation, making a vehicle unrentable
- A PATCH with `minRentalHours > maxRentalHours` silently persisted an impossible booking window
- Added `superRefine` to update schema that checks cross-field constraints only when both fields are present

## Test plan
- [x] 5 new tests: rejects both-rates-null, rejects min > max, allows single-rate null, allows one-rate-with-other-null, allows min-only partial
- [x] All 41 vehicle validator tests pass
- [x] tsc --noEmit passes all three packages
- [x] Biome lint clean

Closes #141

### DIFF
--- a/packages/shared/src/validators/vehicle.ts
+++ b/packages/shared/src/validators/vehicle.ts
@@ -53,7 +53,32 @@ export const createVehicleSchema = vehicleObjectSchema.superRefine((data, ctx) =
   }
 })
 
-export const updateVehicleSchema = vehicleObjectSchema.partial()
+export const updateVehicleSchema = vehicleObjectSchema.partial().superRefine((data, ctx) => {
+  // Only reject when both rates are explicitly present and both null.
+  // A partial update sending only one rate is fine — the other stays unchanged in DB.
+  if (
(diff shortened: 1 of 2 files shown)


jackli921:
Duplicate of #169 (simpler, single-commit version).

Issue #76, opened 2026-04-12
security: Hono API has zero authentication — add JWT middleware + forward from web + partner API key


jackli921:
## Problem

The Hono API in `packages/api/` has ZERO authentication middleware. `createApp()` in `packages/api/src/index.ts:86-112` mounts CORS then routes — nothing in between.

### Confirmed exploits (all require only `curl`, no credentials):

1. `POST /bookings` trusts `renterId` from request body (`routes/bookings.ts:94`). Anyone can create a booking as any user.
2. `POST /vehicles`, `PATCH /vehicles/:id`, `PATCH /vehicles/:id/status`, `DELETE /vehicles/:id` are all wide open (`routes/vehicles.ts:26-108`). A tourist can retire the entire fleet, zero-price every car, or inject phishing URLs into `photos[]`.
3. `GET /threads?userId=<any>` returns every thread for any user id passed in (`routes/messages.ts:12`). Full PII read.
4. `POST /threads/:id/messages` accepts `senderId` from body (`routes/messages.ts:57`). Impersonate any user in any thread.
5. `POST /threads/:id/read` (`routes/messages.ts:74`). Mark any user's threads as read.
6. `PATCH /bookings/:id/status` + `POST /bookings/:id/cancel` have no ownership check. Anyone can cancel anyone else's booking.

The Trip.com server-to-server contract (planned, issue #13) also needs auth — a shared-secret API key per partner.

From the security review (2026-04-11), classified CRITICAL — the ship-blocker.

## Fix (multi-slice)

**Slice A — auth middleware:**
- Add `requireAuth()` Hono middleware that verifies the Auth.js JWT from `Authorization: Bearer <token>` using the same `AUTH_SECRET` the web package signs with. `jose` is edge-safe.
- On success, set `c.set('user', { id, role })`. On failure, 401.
- For partners, accept `X-API-Key` as an alternative that maps to `{ id: 'partner:tripcom', role: 'PARTNER' }`.
- Mount before all routes except `/health`.

**Slice B — web fetchers forward JWT:**
- Create `packages/web/src/lib/api-client.ts::authedFetch(path, init)` that reads the session cookie and adds `Authorization: Bearer <jwt>`.
- Route every server-action fetcher (`lib/bookings.ts`, `lib/messaging.ts`, `lib/dashboard-stats.ts`, `lib/vehicle-api.ts`, etc.) through it.

**Slice C — actor derivation + ownership checks:**
[...]

PR #202, opened 2026-04-12
fix(api): type-safe role validation and fleet-overview STAFF gate

jackli921:
## Summary

- Replace `as UserRole` cast chain with `isValidRole()` type guard
- Add `isAuthUser()` runtime guard to `getUser()` (eliminates bare `as` cast)
- Move per-request `VALID_ROLES` Set to module-level `ALL_ROLES`
- Fleet-overview requires STAFF/ADMIN role (business-internal data)

## Test plan

- [x] 22 auth + message route tests pass
- [x] TypeScript strict mode passes

Refs #76

### DIFF
--- a/packages/api/src/middleware/auth.ts
+++ b/packages/api/src/middleware/auth.ts
@@ -15,6 +15,23 @@ export interface AuthEnv {
   }
 }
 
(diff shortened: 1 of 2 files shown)


jackli921:
Superseded by #198 which already landed type-safe role validation (isValidRole) and fleet-overview STAFF gate.

PR #209, opened 2026-04-12
refactor(shared): remove dead senderId/markReadSchema from message validators

### LINKED ISSUE #203: refactor(shared): make senderId/userId optional in message validators
## Problem

`sendMessageSchema` requires `senderId` and `markReadSchema` requires `userId` in the request body, but the API routes derive both from the JWT `sub` claim and ignore the body values entirely.

This creates a confusing contract: clients must send a field that the server throws away. If a client omits it, the request fails validation even though the server never uses it.

**Current state (after #181, #198):**
- `POST /threads/:id/messages` — route uses `user.id` from JWT, ignores `parsed.data.senderId`
- `POST /threads/:id/read` — route uses `user.id` from JWT, `markReadSchema` not even called anymore

## Fix

Make `senderId` optional in `sendMessageSchema` and `userId` optional in `markReadSchema` (or remove them entirely). Update the shared validator tests accordingly.

**Files:**
- `packages/shared/src/validators/message.ts`
- `packages/shared/tests/validators/message.test.ts`

## Context

Found during code review of #193. The validator change was included in that PR's worktree but didn't make it into the merged #181.

jackli921:
## Summary

- Remove `senderId` from `sendMessageSchema` — routes derive it from JWT
- Remove `markReadSchema` entirely — route doesn't parse body for `/read`
- Remove `MarkReadInput` type export
- Update validator tests to match new contract

Net: -39 lines. Clients no longer need to send a field the server ignores.

## Test plan

- [x] 14 shared validator tests pass
- [x] 14 message route tests pass
- [x] TypeScript strict mode passes for api + web

Closes #203

### DIFF
--- a/packages/shared/src/validators/message.ts
+++ b/packages/shared/src/validators/message.ts
@@ -8,14 +8,8 @@ export const createThreadSchema = z.object({
 })
 
 export const sendMessageSchema = z.object({
-  senderId: z.string().uuid('senderId must be a valid UUID'),
   content: z.string().trim().min(1, 'Message cannot be empty').max(5000),
-})
-
-export const markReadSchema = z.object({
-  userId: z.string().uuid('userId must be a valid UUID'),
 })
 
 export type CreateThreadInput = z.infer<typeof createThreadSchema>
 export type SendMessageInput = z.infer<typeof sendMessageSchema>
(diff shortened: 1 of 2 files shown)


jackli921:
Main already has this fix (landed via another session).

PR #215, opened 2026-04-12
test(api): integration tests for POST /bookings overlap + fix pgErrorCode

### LINKED ISSUE #30: test(api): integration test for POST /bookings overlap → 409
## Context

The in-memory booking repo now mirrors the DB exclusion constraint and throws an error with `code: '23P01'` on overlap (`packages/api/src/repositories/in-memory.ts`), and the `POST /bookings` route translates that to a `409 Conflict` (`packages/api/src/routes/bookings.ts:109-135`). Three new unit tests in `packages/api/tests/routes/bookings.test.ts` cover this behavior at the route level using the in-memory repo.

**Gap:** no integration test proves that the real Postgres path also produces the 409. This session manually verified it by crafting a raw `INSERT` that got rejected with `23P01`, but nothing in CI enforces that the `DrizzleBookingRepository.create` call surfaces the same error code that the in-memory mirror uses.

The risk: if `postgres.js` ever changes how it formats PG error codes (e.g., wraps them in a different property), the in-memory tests would still pass but the real-DB path would fall through to a 500.

## Acceptance criteria

- [ ] New test in `packages/api/tests/integration/bookings.test.ts` (or similar) that:
  - [ ] Creates a CONFIRMED booking via `DrizzleBookingRepository`
  - [ ] Attempts a second overlapping booking on the same vehicle
  - [ ] Asserts the thrown error has `err.code === '23P01'` (or whatever shape `postgres.js` actually produces — capture the real error to find out)
- [ ] Second test via the HTTP route: hits `POST /bookings` twice against the real DB, asserts the second returns `409` with the expected error body
- [ ] Third test covering the CANCELLED case: cancel the first booking, verify the second can be created successfully (proves the exclusion constraint's `WHERE status IN ('CONFIRMED','ACTIVE')` clause is respected end-to-end)

## Notes

- Depends on #NN-integration-tests-in-ci — the new test is only useful if CI actually runs the integration suite.
- The manual verification script from this session (`/tmp/verify-constraint.ts`, now deleted) is a good starting point for the assertions.

jackli921:
## Summary

- Add integration tests proving the booking overlap exclusion constraint works end-to-end against real Postgres
- Fix `pgErrorCode()` to check `err.cause.code` (postgres-js wraps PG errors in `cause`)
- Add unit tests for `pgErrorCode` itself

## Bug found

`pgErrorCode()` only checked `err.code`, but postgres-js puts the PG error code at `err.cause.code`. The real DB path returned **500 instead of 409** on booking overlap. In-memory tests masked this.

## Tests added

| Test | What it proves |
|------|---------------|
| Overlap -> `23P01` | Exclusion constraint fires, pgErrorCode extracts code through `err.cause` |
| Cancel -> rebooking succeeds | `WHERE status IN ('CONFIRMED','ACTIVE')` works |
| Adjacent bookings succeed | Boundary: `startAt === prior effectiveEndAt` allowed |
| HTTP POST -> 409 | Full stack: service catches 23P01, returns 409 |
| Concurrent overlap -> one wins | Race condition handled |
| pgErrorCode unit tests | 8 edge cases |

## Test plan

- [x] 58 integration tests pass
- [x] 193 unit tests pass
- [x] Pre-commit hooks pass

Closes #30

### DIFF
--- a/packages/api/src/pg-errors.ts
+++ b/packages/api/src/pg-errors.ts
@@ -7,11 +7,28 @@ export const PG_ERROR = {
   UNIQUE_VIOLATION: '23505',
 } as const
 
-/** Extract the Postgres error code from an unknown thrown value, or null. */
+/** Extract the Postgres error code from an unknown thrown value, or null.
+ *
+ * Drizzle + postgres-js wraps the raw PostgresError inside `err.cause`,
+ * so the PG error code lives at `err.cause.code`, not `err.code`.
+ * We check both paths so the same helper works with raw PG errors
+ * (in-memory repo) and wrapped drizzle errors (real DB). */
 export function pgErrorCode(err: unknown): string | null {
-  if (err && typeof err === 'object' && 'code' in err) {
-    const code = (err as { code: unknown }).code
+  const code = extractCode(err) ?? extractCode(getCause(err))
+  return code
+}
+
+function extractCode(val: unknown): string | null {
+  if (val && typeof val === 'object' && 'code' in val) {
+    const code = (val as { code: unknown }).code
     return typeof code === 'string' ? code : null
+  }
+  return null
+}
+
+function getCause(val: unknown): unknown {
+  if (val && typeof val === 'object' && 'cause' in val) {
+    return (val as { cause: unknown }).cause
   }
   return null
 }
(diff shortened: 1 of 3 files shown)


jackli921:
Duplicate — work already merged via #214

Issue #285, opened 2026-04-17
security(fleet): vehicle photo upload — content-type spoofing, TOCTOU, delete race


jackli921:
Follow-ups from PR #244 code review. **Impact: stored XSS on origin + silent photo data loss under concurrency.**

## CRITICAL: Content-type spoofing → stored XSS

**File:** `packages/api/src/routes/vehicle-photos.ts:10-13`

`validateFile` trusts the multipart `Content-Type`. Attacker uploads HTML labeled `image/jpeg`, R2 stores and serves it with that contentType → stored XSS on origin.

**Fix:** validate magic bytes before accepting:
- JPEG: `FF D8 FF`
- PNG: `89 50 4E 47 0D 0A 1A 0A`
- WebP: `RIFF....WEBP`
- AVIF: `ftypavif` / `ftypavis` etc.

Reject if Content-Type and bytes disagree. Also set `Content-Disposition: attachment` on R2 PUT or serve from cookie-less subdomain.

## HIGH: TOCTOU on photo count + array write

**File:** `vehicle-photos.ts:29-40`

```ts
// read
const vehicle = await repo.findById(id)
if (vehicle.photos.length >= 10) return fail(...)
// ... R2 put ...
// write
await repo.update(id, { photos: [...vehicle.photos, newUrl] })
```

Two concurrent uploads: both see `length=9`, both pass, both `update(...)` — last write wins, earlier upload orphaned in R2, no DB pointer.

**Fix:** push append into the repo as an atomic op with length guard (SQL `array_append` + `WHERE cardinality(photos) < 10`, retry/409 on violation).

## HIGH: Orphan on partial multi-file upload

**File:** `vehicle-photos.ts:38`

`Promise.all(files.map(storage.put))` — if file 3/5 rejects, files 1-2 are in R2 with no cleanup. The try/catch only runs if `repo.update` throws.

**Fix:** track successful puts, delete them if any put or the subsequent DB write fails.

## HIGH: Index-based delete race

**File:** `vehicle-photos.ts:62-73`

Client passes numeric `:photoIdx`. Concurrent deletes against `[a,b,c]`:
- A reads, deletes idx 0 → saves `[b,c]`
- B reads same `[a,b,c]`, meant to delete `b` (idx 1), saves `[a,c]` (because it wrote based on its stale copy)

Net: `a` resurrected in DB but already deleted from R2.

[...]

PR #290, opened 2026-04-17
security(api): block content-type spoofing in vehicle photo upload

jackli921:
## Summary
Fixes the CRITICAL item from #285 — stored XSS via content-type spoofing on photo upload.

Before: multipart `Content-Type: image/jpeg` was trusted; attacker could upload HTML labeled as a JPEG and R2 would serve it with that content-type, enabling stored XSS on origin.

After: sniff first 16 bytes, verify against magic byte signatures for JPEG/PNG/WebP/AVIF, reject with 415 when bytes don't match the declared MIME (or match nothing).

## Scope
- CRITICAL: content-type spoofing — **this PR**
- HIGH items (TOCTOU, partial-upload orphans, index-based delete race) are scoped for a separate PR

## Test plan
- [x] 11 unit tests on `detectImageType` (valid + spoofing + truncation cases)
- [x] 3 route-level tests: HTML-as-JPEG → 415, PNG-as-JPEG → 415, PNG-as-PNG → 201
- [x] 324 total tests passing
- [x] tsc clean, biome clean

Refs #285

### DIFF
--- a/packages/api/src/lib/magic-bytes.ts
+++ b/packages/api/src/lib/magic-bytes.ts
@@ -1,0 +1,46 @@
+export type ImageType = 'jpeg' | 'png' | 'webp' | 'avif'
+
+/**
+ * Sniff the image type from the first bytes of a file. Returns null if the
+ * bytes don't match any allowed image format. Used to block content-type
+ * spoofing — a client-declared `image/jpeg` header does not guarantee JPEG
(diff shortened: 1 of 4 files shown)


jackli921:
Superseded by the comprehensive photo-security PR that landed in main (#285 was already fully addressed there — magic-byte validation, TOCTOU fix, orphan cleanup, and URL-based delete all in one).

Issue #301, opened 2026-04-17
feat: vehicle class abstraction (renter books class, owner assigns car)


jackli921:
## Summary

Decouple the renter-facing catalog from the owner's physical fleet. Renters browse and book *classes* (e.g. "Compact — 5 seats"); the owner manages individual cars tagged to a class. Inspired by NicoNico Rent-a-Car.

Design: [`docs/plans/2026-04-14-vehicle-class-abstraction.md`](../blob/main/docs/plans/2026-04-14-vehicle-class-abstraction.md)

## Why

1. Renter friction — tourists care about size/seats, not which specific Aqua.
2. Owner inflexibility — a booked car in maintenance currently blocks the booking; with classes, owner can swap within the class.
3. Catalog mismatch — owner wants a clean class-based catalog, not 40 individual cars.
4. 3rd-party aggregators (Trip.com, #13) list by category, not by VIN.

### Phase 1: Schema (additive)
- [ ] Create `vehicle_classes` table (id, name, slug, description, photos, seats, luggageCapacity, transmission, fuelType, dailyRateJpy, hourlyRateJpy, sortOrder, status).
- [ ] Add `classId` to `vehicles` (nullable → backfill → not-null).
- [ ] Add `classId` to `bookings` (nullable → backfill → not-null).
- [ ] Make `bookings.vehicleId` nullable (null = unassigned).
- [ ] Backfill: create classes from existing vehicle data, assign classIds.

### Phase 2: API
- [ ] CRUD routes for vehicle classes (`/api/vehicle-classes`).
- [ ] Availability endpoint per class (count of free cars in class over date range).
- [ ] Booking creation accepts `classId` instead of `vehicleId`.
- [ ] Update fleet overview to group by class.

### Phase 3: Web — renter
- [ ] Catalog page: browse by class (NicoNico-style).
- [ ] Booking flow: select class → pick dates → confirm.
- [ ] Remove individual vehicle browsing from renter side.

### Phase 4: Web — owner
- [ ] Vehicle class management UI (CRUD).
- [ ] Fleet view grouped by class with class-level stats.
- [ ] Booking detail: vehicle assignment dropdown (available cars in class).

## Open questions (resolve before Phase 1)

1. Pricing — class-level default with per-vehicle override? (proposed: yes)
2. i18n for class descriptions — next-intl keys vs. stored translated text?
3. Auto-assignment timing — at booking time vs. deferred to owner? (proposed: deferred for MVP)
[...]

PR #317, opened 2026-04-17
feat(api): public class catalog + availability endpoint

### LINKED ISSUE #309: feat(api): public class catalog + availability-per-class endpoint
Part of #301. Backend-only.

## Scope
- Make `GET /vehicle-classes` + `GET /vehicle-classes/:slug` public (no auth)
- Add `GET /vehicle-classes/:slug/availability?from=&to=` returning count of free cars in class for date range
- Rate-limit the public endpoints

## Depends on
- #251 (merged)

jackli921:
Closes #309
Refs #301

## Summary
- Makes `GET /vehicle-classes` and `GET /vehicle-classes/by-slug/:slug` unauthenticated so the renter catalog can render without login.
- Adds `GET /vehicle-classes/:slug/availability?from=&to=` returning `{ totalCars, availableCars, sampleAvailableVehicleIds }`, reusing the existing availability repo so `effectiveEndAt` + buffer logic stays single-sourced.
- Stacks a per-IP `PUBLIC_CATALOG_LIMITER` (30/min) on top of the global limiter for the three public endpoints.
- `GET /vehicle-classes/:id` + all mutations keep requiring auth + STAFF role via per-route `requireAuth` inside `createVehicleClassRoutes`.

## Design notes
- Split availability into a dedicated `VehicleClassAvailabilityService` rather than expanding `VehicleClassService`; the CRUD service's single responsibility is class records, availability is a cross-aggregate query.
- `sampleAvailableVehicleIds` capped at 5 to bound payload size for the unauth endpoint.
- `requireAuth()` now short-circuits if upstream already set `c.get('user')`, so public + protected routes compose cleanly in one Hono instance.

## Test plan
- [x] `bun run test` — 400 API + 315 web + 17 shared = 732 green
- [x] `bunx tsc --noEmit` on packages/api — clean
- [x] `bun run lint` — clean (pre-existing unrelated warning only)
- [x] `bun run --filter @kuruma/api lint:boundaries` — OK
- [x] New test files:
  - `tests/routes/public-catalog.test.ts` (19 cases: unauth list/slug/availability, auth-required mutations, edge cases incl. no vehicles / all booked / partial overlap / RETIRED excluded / 404 / 400)
  - `tests/routes/public-catalog-ratelimit.test.ts` (4 cases: 429 after cap, shared bucket across three endpoints, per-IP isolation, local-dev bypass)
- [ ] Manual smoke against `wrangler dev` once deployed (PUBLIC_CATALOG_LIMITER binding is declared in wrangler.toml)


jackli921:
Superseded by #320 — same issue (#309), smaller diff, aggregated SQL count instead of in-memory count, more test coverage. `sampleAvailableVehicleIds` dropped as YAGNI; can be added as a follow-up if a consumer surfaces.

PR #322, opened 2026-04-17
feat(api): booking-by-class plumbing

### LINKED ISSUE #308: feat(api): booking-by-class plumbing (classId on bookings, nullable vehicleId)
Part of #301. Backend-only. No UI change.

## Scope
- Add `classId` (FK vehicle_classes, not null) to `bookings` table
- Make `bookings.vehicleId` nullable (null = unassigned, owner assigns later)
- Update booking creation API: accept `classId`; `vehicleId` optional
- Update availability/conflict checks to handle null vehicleId (reserve against class pool)
- Backfill existing bookings with classId derived from `vehicles.classId`

## Out of scope
- UI changes (covered by later slices)
- Availability-per-class endpoint (#301b)

## Depends on
- #251 (merged) — provides classes table + vehicles.classId

jackli921:
Closes #308
Refs #301

## Summary
- Adds `classId` (FK → `vehicle_classes`, NOT NULL) to `bookings` and makes `bookings.vehicleId` nullable. Renters now book a class; the owner assigns a specific vehicle later.
- `createBookingSchema` accepts `classId` as required and `vehicleId` as optional.
- `BookingService` verifies the class exists and is active, and — when a vehicleId is supplied — that the vehicle belongs to the chosen class. Pricing falls back from vehicle override → class rate.
- Unassigned bookings (`vehicleId=null`) skip per-row overlap exclusion; class-pool capacity is the concern of the availability endpoint (#301b).
- Migration 0024 adds `classId`, backfills from `vehicles.classId`, enforces NOT NULL with a diagnostic `RAISE` if any row is unbackfilled, and drops NOT NULL on `vehicleId`.

## Implementation notes
- Migration is hand-authored because `drizzle/meta` has duplicate snapshot ids on main (0021/0022) which confuse `drizzle-kit generate`. The snapshot file is generated normally and reflects the correct final schema; only the SQL text is by hand. Follow-up: clean up the duplicate snapshot ids.
- New DI dependency: `BookingService` takes a `VehicleClassRepository`. `lookupClass`/`lookupVehicle` throw if their repo is missing, rather than silently masquerading as "not found".
- Fleet-overview query now skips rows where `vehicleId IS NULL` (class-only bookings don't aggregate into any specific vehicle).

## Test plan
- [x] `bun run test` — 873 tests green (171 shared + 387 api + 315 web)
- [x] `bun run tsc --noEmit` — clean across all 3 packages
- [x] `bun run lint` + `lint:boundaries` — clean
- [x] `bun run db:verify` — schema ↔ snapshot + journal ↔ disk in sync
- [x] New service tests cover: class-only booking, vehicle-class mismatch, unknown class, unknown vehicle, archived class, no conflict for unassigned bookings
- [ ] Apply migration against a real DB branch and confirm backfill populates every row (requires DATABASE_URL — CI job `db-drift` covers schema, manual verification recommended for backfill before deploy)
- [ ] Web UI changes covered by later slices (out of scope for #308)


jackli921:
## Code review — PR #322 (booking-by-class plumbing)

Reviewed priority areas: migration safety, exclusion invariant, class/vehicle mismatch, pricing fallback, snapshot id chain, and CallerContext. Overall this is a well-structured, low-risk plumbing PR. Findings below, ordered by severity.

---

### MEDIUM

**[MEDIUM] Exclusion constraint relies on implicit NULL semantics (no partial index)**
- File: `drizzle/0006_exclusion_constraint.sql` (unchanged) + behavior depended on by `drizzle/0024_booking_class_fk.sql`
- The existing constraint is `EXCLUDE USING gist ("vehicleId" WITH =, tstzrange(...) WITH &&) WHERE (status IN ('CONFIRMED','ACTIVE'))`. With `vehicleId` now nullable, safety relies on Postgres treating `NULL = NULL` as "not distinct" — i.e. rows with a NULL vehicleId never participate in the `=` predicate and therefore never collide. This is correct (documented in `in-memory/booking.ts:119`), but it is implicit. Two risks: (a) future maintainers adding `nulls not distinct`–style clauses could flip the semantics; (b) GiST still indexes the NULL rows, wasting space on a growing null-heavy table.
- Fix (non-blocking, follow-up): tighten the constraint to `... WHERE (status IN ('CONFIRMED','ACTIVE') AND "vehicleId" IS NOT NULL)`. Makes the intent explicit and keeps the index lean. Ship in a 0025 follow-up so this PR stays focused on plumbing.

```
Learn: Implicit NULL semantics in exclusion constraints
Postgres gist EXCLUDE with `WITH =` treats two NULLs as non-matching, so NULL rows naturally don't conflict.
Relying on that silently couples correctness to an SQL subtlety; a future schema tweak (or a different DB) can flip the behavior without warning.
Heuristic: if a nullable column participates in an EXCLUDE/UNIQUE, make the null-handling explicit via `WHERE col IS NOT NULL` or `NULLS NOT DISTINCT`.
```

**[MEDIUM] Class/vehicle consistency not enforced after booking creation**
- File: `packages/api/src/services/booking.ts:278` (`resolveClassAndVehicle`)
- `resolveClassAndVehicle` checks `vehicle.classId === input.classId` at creation time, but nothing prevents the owner from later reassigning that vehicle to a different class (e.g. Compact → SUV). The booking row's `classId` would then disagree with `bookings.vehicle.classId`, and pricing/availability derived downstream could diverge.
- Fix (business decision, surface to product): decide the rule — either (a) forbid reclassifying a vehicle that has active/future bookings; (b) snapshot the class rate/name onto the booking at creation (denormalize); or (c) accept the drift and document it. A `Learn` for the team:

```
Learn: Snapshot vs live reference (DDD / aggregate invariants)
A booking's class is currently a live FK, so any mutation to the vehicle's class retroactively changes what the renter booked.
At rental scale this shows up as support tickets ("I booked a Compact and got charged SUV"); the longer the window, the worse.
Heuristic: data that represents a commitment at time T (price, class, terms) should be snapshotted onto the commitment row, not referenced live.
```

**[MEDIUM] Hand-authored migration bypasses drizzle-kit's invariant checks**
- File: `drizzle/0024_booking_class_fk.sql` + `drizzle/meta/0024_snapshot.json`
- The PR body explains this is necessary because 0021/0022 snapshots share an id on `main` (confirmed: both show `1c559150-9101-42a7-89c7-a9a2da918740`). The 0024 chain is correctly rooted at 0023 (`prevId: 051d9009...` → matches 0023's id). Migration logic itself is sound: add nullable → backfill → RAISE if incomplete → add FK → SET NOT NULL → drop NOT NULL on vehicleId. Diagnostic RAISE correctly runs **before** `SET NOT NULL` (lines 27–36), so an unbackfillable row aborts with a readable message instead of a cryptic 23502.
- Fix (follow-up, not this PR): file an issue to regenerate 0021/0022 snapshot ids (or mark the stale ones) so drizzle-kit can resume auto-generating. Hand-authored migrations are a one-off escape hatch, not a workflow.

---

### LOW

**[LOW] `resolveClassAndVehicle` return type blends success + error shapes**
- File: `packages/api/src/services/booking.ts:267`
- Return type is `(CreateBookingResult & { ok: false }) | { ok: true; bufferMinutes; totalPrice }`. The `& { ok: false }` intersection works but is awkward; a cleaner discriminated union would be `{ ok: false; error: CreateBookingResult & { ok: false } } | { ok: true; ... }` or just returning `CreateBookingResult` directly with the buffer/price on success. Not a bug — readability only.

**[LOW] `NO_RATES_SET` path is reachable but effectively dead**
- File: `packages/api/src/services/booking.ts:333`
- Vehicle classes have a `vehicle_classes_pricing_at_least_one` CHECK, and vehicle override can only *fall back* to the class rate. So `dailyRateJpy ?? hourlyRateJpy` can only both be null if the class row somehow violates its check (impossible). The defensive branch is fine; worth a comment noting it's defense-in-depth.

---

### What's solid

- Migration is idempotent in the useful sense: if the backfill UPDATE already populated rows, re-running the UPDATE is a no-op (guarded by `classId IS NULL`), and the RAISE-then-NOT-NULL path is correct.
- In-memory repo (`getConflictingBookings`, `create`) faithfully mirrors the Postgres NULL-vehicleId semantics; tests exercise the same path.
- Pricing fallback (override → class rate) is correct and covered by tests (`bookings.test.ts`, new cases for class-only booking + mismatch + archived class).
- Snapshot id chain from 0023 → 0024 is valid. The 0021/0022 duplication is pre-existing on main and not introduced here.
- `lookupClass` / `lookupVehicle` throwing on missing DI rather than returning null is the right call — silent "not found" would have masked DI bugs for months.
- No concurrent-write surface in this PR (no assignment endpoint yet); race concerns land in a future slice.
- Route layer correctly threads `classId` through and keeps ownership scoping in the repo via `CallerContext`. No regressions.

---

### Not blocking, but worth filing follow-ups

1. Tighten exclusion constraint with explicit `WHERE "vehicleId" IS NOT NULL` (MEDIUM above).
2. Decide snapshot-vs-live policy for booking↔class consistency before an assignment endpoint lands (MEDIUM above).
3. Regenerate 0021/0022 snapshot ids so drizzle-kit auto-generate works again.
4. When the assignment endpoint lands (#301b-ish), it will need `SELECT ... FOR UPDATE` on the booking row **or** rely on the exclusion constraint firing during the UPDATE. Flag at planning time.

Overall: looks good to merge once the class/vehicle consistency policy is decided (can be a doc update in this PR). Migration is safe to apply.

jackli921:
Retriggering CI — local typecheck is clean but CI showed a stale type error on in-memory/booking.ts:110. Re-running.

PR #338, opened 2026-04-17
feat(web): renter catalog — browse by vehicle class

### LINKED ISSUE #310: feat(web): renter catalog — browse by vehicle class
Part of #301. Renter-facing UI.

## Scope
- Rebuild `/vehicles` list page as class catalog (NicoNico-style)
- Show class hero photo, name, seats, transmission, daily rate
- Class detail page `/vehicles/classes/:slug` with description + photos
- Remove individual-vehicle browsing from renter nav

## Depends on
- #301b (public catalog API)

jackli921:
Closes #310
Refs #301

## Summary

- Rebuilds `/vehicles` as a NicoNico-style class catalog: hero photo, name, seats, transmission, daily rate, linking to the class detail page.
- Adds `/vehicles/classes/[slug]` detail page with description, photo gallery, specs, and a booking CTA keyed by class slug.
- Removes the old renter `/vehicles/[id]` detail page and the orphaned `components/catalog/ClassCard.tsx` left behind from a prep PR.
- Adds `fetchClassBySlug()` calling the public `GET /vehicle-classes/by-slug/:slug` endpoint (no auth, from #317).
- Adds `catalog` + `catalog.detail` i18n namespaces in en/ja/zh.
- Owner routes at `/manage/*` and the `getVehicleById` lib helper stay intact.

## Test plan

- [x] `bun run test` — 945 passed (web 359 / api 417 / shared 169), +9 new tests
- [x] `bun run --filter @kuruma/web typecheck`
- [x] `bun run lint` (biome + file-size + module-boundaries)
- [ ] Manual: `/vehicles` renders the class grid
- [ ] Manual: `/vehicles/classes/compact` renders detail view
- [ ] Manual: unknown slug 404s via `notFound()`
- [ ] Manual: logged-out visitor can browse without redirect


jackli921:
## Code review — PR #338

Scope: renter catalog UI. Priority areas all covered. No CRITICAL findings. One HIGH (brittle 404 detection), a few MEDIUM, a couple LOW.

---

### HIGH

**[HIGH] Slug not-found detection relies on exact error message string**
`packages/web/src/modules/classes/api.ts:66`
```ts
if (e instanceof Error && e.message === 'Vehicle class not found') return null
```
Any change to the API's 404 copy (wording, casing, i18n, generic fallback) silently turns a 404 into a thrown error and triggers `error.tsx` instead of `notFound()`. Same pattern already exists in `fetchClassById` — both are fragile.
Fix: branch on HTTP status instead of message text. Extend `unwrap<T>` (or add `unwrapOr404`) to return `null` on 404 and throw on other non-2xx, then `fetchClassBySlug` becomes a one-liner.

```
Learn: String-matched error branching
Control flow keyed on a human-readable message couples the caller to copy the server owns.
At scale the message drifts (i18n, typo fix, generic fallback) and the branch silently stops firing.
Heuristic: branch on status codes or error classes; strings are for humans, not for `if`.
```

---

### MEDIUM

**[MEDIUM] Type assertion instead of proper narrowing**
`packages/web/src/modules/classes/components/ClassCatalogCard.tsx:57`
`formatJpy(vehicleClass.dailyRateJpy as number)` — the `as number` is only needed because `hasPrice` narrows a different binding. Violates `typescript.md` (no `as` without type guard).
Fix: inline the check — `{vehicleClass.dailyRateJpy != null && (<p>…{formatJpy(vehicleClass.dailyRateJpy)}…</p>)}`. Same shape `ClassDetailView.tsx` already uses.

**[MEDIUM] Unused i18n keys shipped across all 3 locales**
`catalog.viewClass`, `catalog.detail.description`, `catalog.detail.notFound`, `catalog.detail.notFoundDescription` are defined in en/ja/zh but never read by any component. `notFound*` in particular reads like it was meant for a custom 404 page in `vehicles/classes/[slug]/not-found.tsx` that didn't land.
Fix: drop them, or add the `not-found.tsx` that consumes them (nicer renter UX than the generic `[locale]/not-found.tsx`).

**[MEDIUM] Duplicate `catalog` and `catalog.detail` keys**
`priceFrom`, `perDay`, `seats`, `auto`, `manual` exist in both `catalog.*` and `catalog.detail.*` with identical values in en/ja/zh. Future copy edits risk drifting between catalog card and detail view.
Fix: keep one source (`catalog.*`) and read from the parent namespace in the detail component, or factor shared strings into `catalog.shared.*`.

---

### LOW

**[LOW] Catalog fetch runs on every request**
`fetchClasses` / `fetchClassBySlug` use plain `fetch` via `hc`. Next 15+ server-component default is no caching. For a public, rarely-changing catalog, a short `next: { revalidate: 60 }` (or a `revalidateTag('vehicle-classes')` on owner-side mutations) would cut API load visibly once traffic scales. Not a blocker for this PR — raise a follow-up.

**[LOW] `` instead of `next/image`**
Both new components use raw ``. Consistent with the deleted file, so not a regression, but catalog hero photos are exactly the place where `next/image` pays off (LCP, responsive srcset). Worth a follow-up.

---

### Verified clean

- Public endpoint behavior: `fetchClassBySlug` passes no token to `createApiClient()`; `createApiClient` only attaches `Authorization` when a token is truthy. No auth leakage on the public path. Test `api.test.ts:109` explicitly asserts the header is absent.
- SSR/SEO: both `vehicles/page.tsx` and `vehicles/classes/[slug]/page.tsx` are server components. No `'use client'`. Headings sensible (`h1` page title → `h2` per-card / `h2` Specifications).
- i18n coverage: top-level and `detail.*` key sets are identical across en/ja/zh (diff confirmed). No hard-coded user-visible English strings in the new components.
- No broken links: no remaining references to `/vehicles/[id]` in renter routes. `/manage/vehicles/:id` owner routes untouched.
- 404 path: `fetchClassBySlug` returns `null` → page calls `notFound()` → Next routes to `[locale]/not-found.tsx`. Generic API error bubbles to `[locale]/error.tsx`. Both exist.
- Rate limiting: `publicCatalogLimiter` stacks on `/vehicle-classes*` in the API, so the new public fetches already carry a tighter per-IP budget (shipped in #317).
- A11y: hero and thumbnail `` have `alt={vc.name}`; photo `<section>` has `aria-label`; `<Link>` wrappers contain text content or wrap image+heading (accessible names fine).

No action required for the clean items — just confirming the priority checklist.

PR #341, opened 2026-04-17
fix(db): restore 0023 journal monotonicity + detect future drift

jackli921:
## Summary
- Bump `0023_add_phone_to_users` `when` to `max(previous) + 1` so drizzle-kit migrate applies it on fresh DBs (CI + new environments). Prior `when` was ~3.3 hours earlier than 0022, triggering drizzle's silent-skip behaviour.
- Add `db:verify` Check 3 — journal `when` monotonicity. Catches the bug class at commit time instead of at prod crash time.
- Document the 2026-04-17 incident in `CLAUDE.md` (under the drizzle migrations section).

## Root cause
drizzle-kit compares each journal entry's `when` field to the last-applied migration's `when`. Entries with an earlier timestamp are treated as "already applied" and skipped, even when their hash is absent from `__drizzle_migrations`. The `when` fell out of order after a rebase/cherry-pick of 0023 onto a main that had since merged later migrations (0020–0022).

## Impact
- Before: booking API returned `column "phone" does not exist`; PR #322 db-drift job failing with the same symptom on a fresh DB.
- After: fresh-DB `db:migrate` applies 0023. Prod (already at the right schema via manual recovery) is untouched — drizzle's hash guard still recognises 0023 as applied.

## Test plan
- [x] Drop + recreate `kuruma_test` locally
- [x] `bun run db:migrate` applies all 24 migrations
- [x] `bun run db:verify` — 4/4 checks pass
- [x] `\d users` confirms `phone` column exists
- [x] `bun run --filter @kuruma/api test:integration` — 74/74 pass
- [ ] CI on this PR is green
- [ ] After merge: rebase #322 and bump 0024's `when` (same root cause, different migration)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,6 +98,11 @@ bun run db:verify   # must show 3 green checks
 
 **`bun run db:verify`** checks: schema-snapshot sync, journal-disk sync, journal-DB sync. CI enforces via `db-drift` job. Run before any commit touching `schema.ts` or `drizzle/`.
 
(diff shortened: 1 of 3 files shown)


jackli921:
Superseded by the new PR; this branch could not be rebased onto the latest main without a force push.

Issue #330, opened 2026-04-17
perf(db): add index on vehicles.classId FK + lint for unindexed FKs


jackli921:
**From architect audit (main @ ef1dd8c).**

`packages/shared/src/db/schema.ts:113` — `classId` FK added in migration 0019, but migration 0010 (the FK-index audit) predates it. Any query filtering vehicles by class will sequential-scan.

## Why it bites at scale
At 40–50 vehicles it's invisible. At 2000+ and a public catalog that filters on `classId` (incoming via #301/#317 slices), it's a hot query doing seq scans. Same pattern as the issue #88 incident.

## Learn: FK Added After the Index Audit
Each new FK needs its own index; one-time audits don't cover future migrations.
Every new `.references()` is a future table scan unless paired with an index in the same migration.
**Heuristic:** every `.references(...)` needs a matching index in the SAME migration — enforce in CI, not in retrospect.

## Fix
Migration: `CREATE INDEX idx_vehicles_classId ON vehicles (classId);`. Add a linter: scan `schema.ts` for `.references()` without a corresponding `.index()` or explicit migration.

jackli921:
Superseded by PR #348 (journal `when` fix) and issue #349 (snapshot collision fix). The FK index migration itself is blocked on #349 — once drizzle-kit generate works again, we can declare the indexes in schema.ts and let the generator produce a clean migration. Unlabeling as in-progress; pick up after #349 lands.

PR #348, opened 2026-04-19
fix(db): monotonic journal `when` + db:verify guard against silent-skip

jackli921:
## Why

`drizzle-orm`'s pg migrator applies a migration iff `lastDbMigration.created_at < entry.when`. Cherry-picking migrations left 0023 and 0024 with `when` values **older than** 0022's, so incremental prod deploys silently skipped them. Fresh CI missed it because `lastDbMigration` is null on a clean DB and everything applies. That's the 2026-04-17 `column "phone" does not exist` 15-min outage.

## What

- **Bump 0023 / 0024 `when`** to `max(predecessor) + 1` so the next incremental deploy applies them (or, if they were recovered manually, correctly skips them — the delta is minimal enough that any recovery timestamp ≥ these values still satisfies drizzle's skip condition).
- **Add check 1.5 to `db:verify`**: every entry's `when` must be strictly greater than the previous. Fails CI via the `db-drift` job if violated. Kills the entire class of cherry-pick-breaks-prod bugs at commit time.
- **Update the CLAUDE.md gotcha** to say the monotonic check is now the primary guard.

## Learn: Silent Ordering Bug in Migration Journal
`drizzle-orm` compares `when` (journal) to `created_at` (DB) as the "apply-me" condition; an out-of-order `when` is indistinguishable from "already applied" from a fresh CI's perspective. **Heuristic:** any ordering-sensitive metadata that's hand-authored in a monorepo (migration `when`, schema version numbers) needs a pre-commit invariant check — rebase/cherry-pick will violate it otherwise.

## Test plan

- [x] `bun run db:verify` green locally
- [x] Regression test: temporarily revert 0023's `when` → check fires with the right message (see commit body)
- [ ] CI db-drift runs `db:migrate` against fresh postgres + `db:verify`
- [ ] Confirm prod DB state before deploy (see runbook)

## Out of scope / follow-ups

- **drizzle-kit generate exits 0 despite snapshot-chain collision** (0021 and 0022 share an `id`). Separate issue.
- **FK index enforcement lint** (#330) — deferred.
- **Root cause — index-only migrations without `.index()` in schema.ts** — meta issue, needs repo-wide fix.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,7 +96,13 @@ bun run db:verify   # must show 3 green checks
 
 **Hand-written migrations:** use `bun run db:generate --custom --name <name>`, never drop raw `.sql` into `drizzle/`.
 
-**`bun run db:verify`** checks: schema-snapshot sync, journal-disk sync, journal-DB sync. CI enforces via `db-drift` job. Run before any commit touching `schema.ts` or `drizzle/`.
+**`bun run db:verify`** checks: schema-snapshot sync, monotonic journal `when`, journal-disk sync, journal-DB sync. CI enforces via `db-drift` job. Run before any commit touching `schema.ts` or `drizzle/`.
+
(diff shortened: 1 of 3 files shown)


jackli921:
Duplicate of #342 (which landed ~1 hour before this PR was opened). Closing.

Issue #407, opened 2026-06-03
feat(marketplace): GET /operators + web admin operator picker (hard gate before operator #2)


jackli921:
## Why this is a gate, not a backlog item

#401 removed the hardcoded Best-Car-Rental operator fallback. Write-path operator resolution (`resolveOperatorIdForWrite`, `tenancy.ts`) is now count-gated:

1. `OPERATOR_*` callers -> own tenant (from JWT)
2. non-operator (`PLATFORM_ADMIN` / legacy `STAFF`/`ADMIN`) -> explicit body `operatorId`, else the sole operator if exactly one exists, else **422** (`OperatorRequiredError`).

The web `/manage/{vehicles,classes}` create forms are STAFF/ADMIN-gated and carry **no** `operatorId` in the JWT, so today they rely entirely on sole-operator inference. The moment a second `operators` row is seeded, every admin "create vehicle / create class" returns a bare 422 with no UI affordance to choose an operator. **This blocks operator #2 onboarding.**

Code + architect review on #401 agreed: the 422 fail-closed is correct, but it is only humane once the UI can satisfy it.

## Scope

- [ ] `GET /operators` API route — list operators (admin/platform scope; `OPERATOR_*` sees only its own). Returns id + name + slug.
- [ ] Web: operator picker on the vehicle-create + vehicle-class-create forms; send the chosen `operatorId` in the POST body. Hidden when exactly one operator exists (preserve today's one-click flow).
- [ ] Map the 422 `OperatorRequiredError` to an inline form error prompting operator selection (not a raw toast).
- [ ] Retire sole-operator inference once operator #2 exists: require an explicit `operatorId` for non-operator writes (closes the LOW TOCTOU between `findSoleId()` and the insert noted in #401 review).
- [ ] (optional) Observability: log at the resolver when `findSoleId()` returns null/ambiguous, so the single->multi regime change is visible in telemetry rather than via a support ticket.

## Refs

Blocks operator #2 onboarding (Epic #385). Follows #401, #400, #397.

jackli921:
Merged to marketplace-pivot via PR #433 (squash 5960e1c).

Issue #395, opened 2026-05-29
feat(marketplace): operator-scope VehicleClassRepository (#386 follow-up)


jackli921:
## Context
Deferred from slice 1 tenancy foundation (#386). In slice 1, `vehicle_classes.operatorId` was made NOT NULL with an FK to `operators`, and the seed/test harness stamps the transitional Best Car Rental operator. However the **VehicleClassRepository is not yet operator-scoped**: reads do not filter by `operatorReadScope(ctx)` and writes do not resolve `operatorId` via `resolveOperatorIdForWrite`.

This mirrors what was done for `VehicleRepository` in slice 1 (commit b93285b + Phase 3 pt2).

## Scope
- Drizzle + in-memory `VehicleClassRepository`:
  - Reads (`findAll`/`findById`/etc.): filter by `operatorReadScope(ctx)` — `operator` kind scopes to `ctx.operatorId`, `none` fails closed (empty/undefined), `all` bypasses.
  - Writes (`create`): stamp `operatorId` via `resolveOperatorIdForWrite(ctx, data.operatorId)`.
- Remove the direct `operatorId` stamp workaround in `tests/integration/setup.ts` `seedVehicleClass` once the repo resolves it (keep harness green).
- Add a cross-operator vehicle-class isolation integration test (mirror `tenancy-isolation.test.ts`).

## Acceptance
- An `OPERATOR_OWNER` scoped to tenant A sees only A's classes; a tenant-less OPERATOR_* sees none.
- 93+ integration tests stay green; unit tests green.

Ref: docs/plans/2026-05-25-marketplace-mvp-proposal.md §5/§6; epic #386.

jackli921:
## Sharpened from the cloud review of #398 (2026-06-02)

The pre-merge review of #398 confirmed `VehicleClassRepository` has **no operator scoping at all** (`findAll`/`findById`/`findBySlug` take no `ctx`), and flagged two concrete leaks plus a cross-tenant write. #398 sealed the **vehicle** boundary; these are the **class** half and were intentionally kept here. Exact scope:

1. **Protected `GET /vehicle-classes/:id` is unscoped** (`routes/vehicle-classes.ts:74-78` → `service.findById(id)`). It's behind `requireAuth()` only, so any authenticated `OPERATOR_*` can read another tenant's class. Live read-leak the moment operator #2 exists.
[...]

PR #408, opened 2026-06-03
feat(marketplace): drop BCR operator fallback — count-gated resolution + complete FK→422 (#401, #400)

### LINKED ISSUE #400: feat(api): map PG 23503 (FK violation) to 422 in vehicles route
Follow-up from #395 / PR #399 (composite FK seal).

The composite FK `vehicles(operatorId,classId) -> vehicle_classes(operatorId,id)` now rejects a cross-tenant `classId` at the DB. The vehicles route currently lets that surface as a raw error (500). Map `pgErrorCode(err) === '23503'` to a 422 in both `POST /vehicles` and `PATCH /vehicles/:id` catch blocks (the route already branches on `PG_ERROR.UNIQUE_VIOLATION`; add `FOREIGN_KEY_VIOLATION: '23503'` to `PG_ERROR`).

Deferred from #395 because route tests use in-memory repos, which cannot exercise the DB-only FK — covering the mapping cleanly needs a Drizzle-backed route integration test.

Low priority: the operator UI only offers a tenant's own classes, so this is defense-in-depth, not a live bug. The DB seal itself (the actual security boundary) is already in place and tested in `tenancy-isolation.test.ts`.

### LINKED ISSUE #401: feat(marketplace): drop resolveOperatorIdForWrite Best-Car-Rental fallback
Fast-follow from #397 (operator write-scope). Deferred there to keep the PR small and avoid changing create contracts/UI.

## Problem
`resolveOperatorIdForWrite(ctx, inputOperatorId)` (`packages/api/src/tenancy.ts`) resolves a non-operator caller (PLATFORM_ADMIN / legacy STAFF/ADMIN) to `inputOperatorId ?? BEST_CAR_RENTAL_OPERATOR_ID`. The vehicle/class create routes call it with **no** `inputOperatorId`, so every legacy-admin write is silently stamped with the Best Car Rental tenant. Safe today (single operator), but **misattributes writes to the default tenant the moment operator #2 data exists** — see memory `project_operator-2-gate`.

Operators themselves are now correct (#397): `OPERATOR_*` writes stamp their own `operatorId`. This issue is only about the **non-operator** (platform/legacy) create path.

## Proposed fix
A non-operator caller must supply an explicit target operator; no silent default.
- Add optional `operatorId` to `createVehicleSchema` + `createVehicleClassSchema` (`packages/shared/src/validators/` — danger zone, affects api + web).
- Routes pass `parsed.data.operatorId` into `resolveOperatorIdForWrite(ctx, …)`.
- Reject non-operator writes that omit `operatorId` once 2+ operators exist (or unconditionally), instead of defaulting to BCR.
- Update the web admin create UI to send the chosen operator.

## Gate
Must land **before operator #2 data is created** (functional misattribution, not a security leak — the #399 seal already closed the cross-tenant leak). Coordinate with #400 (PG 23503->422) as the vehicles-route hardening pass.

Refs #397, #400.

jackli921:
## Summary

Removes the hardcoded **Best-Car-Rental** default operator in write paths and replaces it with **count-gated** operator resolution, so a non-operator write can never be silently misattributed once operator #2 exists. Also completes the #400 FK→422 mapping across both vehicle and vehicle-class writes.

## What shipped

**#401 — drop the fallback (`tenancy.ts`)**
`resolveOperatorIdForWrite` is now async + count-gated. A non-operator caller (`PLATFORM_ADMIN` / legacy `STAFF`/`ADMIN`) resolves the target operator by: explicit body `operatorId` → else the sole operator if exactly one exists → else **422** (`OperatorRequiredError`). `OPERATOR_*` callers are unchanged — locked to their own tenant, fail-closed. The hardcoded default is gone.
- `OperatorLookup` + `ResolveWriteOperatorId` (function injection); wired once in `index.ts`, injected into vehicle + class routes (routes never import a repo).
- `OperatorRepository.findSoleId()` on both impls. Vehicle repos no longer self-resolve — route resolves, repo trusts.
- Validators: optional `operatorId` as `z.string().min(1)` (text ids like `op_best_car_rental`, **not** uuid; DB FK is the existence guard).

**#400 — FK→422 mapping (now complete)**
- `POST/PATCH /vehicles`: a `23503` was mapped to a flat "Invalid vehicle class". vehicles carries **two** FKs (composite `operatorId,classId` + single `operatorId`); now branches on the constraint name so a bad operatorId reports "Invalid operator".
- `POST /vehicle-classes`: previously had **no** FK catch — an unknown `operatorId` returned a raw **500** (pages on-call). Now caught → **422 "Invalid operator"**.
- `pg-errors`: `pgConstraintName()` helper + `VEHICLES_CLASS_FK` constant.

## Review

code-reviewer + architect both returned **SHIP**, no CRITICAL/HIGH. Tenant isolation holds (OPERATOR_* can't write cross-tenant; PATCH can't re-tenant), DI boundaries intact, tests mutation-resistant. The convergent finding — incomplete/inconsistent FK→422 mapping — is fixed in this PR (commit `9cacfff`). The LOW TOCTOU on sole-operator inference is deferred to the operator-#2 gate.

## Test plan

- [x] `557` unit/route tests pass
- [x] `115` integration tests pass (was 113; +2 real-PG: unknown operatorId on each route)
- [x] api + web `tsc --noEmit` clean
- [x] import boundaries OK, biome clean (pre-commit enforces all)

## Follow-up

**#407** — `GET /operators` + web admin operator picker. **Hard gate before operator #2 onboarding**: the `/manage/{vehicles,classes}` forms carry no `operatorId` in the JWT and rely on sole-operator inference until the picker lands; the day a 2nd operator is seeded, admin creates 422 until #407 ships.

Closes #401
Closes #400

### DIFF
--- a/packages/api/src/error-handlers.ts
+++ b/packages/api/src/error-handlers.ts
@@ -1,6 +1,6 @@
 import type { Hono } from 'hono'
 import { HTTPException } from 'hono/http-exception'
-import { ForbiddenError } from './middleware/auth'
+import { ForbiddenError, OperatorRequiredError } from './middleware/auth'
 
 export function setupGlobalHandlers(app: Hono): void {
   app.onError((err, c) => {
@@ -16,6 +16,11 @@ export function setupGlobalHandlers(app: Hono): void {
     // clients see a policy denial, not a server outage.
     if (err instanceof ForbiddenError) {
       return c.json({ success: false, error: 'Forbidden' }, 403)
+    }
+    // A non-operator write that named no target operator and could not be
+    // inferred (zero or 2+ operators). Well-formed but unprocessable (#401).
+    if (err instanceof OperatorRequiredError) {
+      return c.json({ success: false, error: err.message }, 422)
     }
     console.error(
       JSON.stringify({
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -84,6 +84,7 @@ import { VehicleClassService } from './services/vehicle-class'
 import { VehicleClassAvailabilityService } from './services/vehicle-class-availability'
 import { VehicleDetailService } from './services/vehicle-detail'
 import { VehiclePhotoService } from './services/vehicle-photo'
+import { type ResolveWriteOperatorId, resolveOperatorIdForWrite } from './tenancy'
 
 export function createApp(overrides?: {
   vehicleRepo: VehicleRepository
@@ -311,6 +312,10 @@ export function createApp(overrides?: {
   const fleetOverviewService = new FleetOverviewService(fleetOverviewRepo)
   const vehicleDetailService = new VehicleDetailService(vehicleDetailRepo)
   const operatorService = new OperatorService(operatorRepo)
+  // #401: bind the write-operator resolver to the concrete operator lookup, so
+  // the write routes resolve the target tenant without importing a repository.
+  const resolveWriteOperatorId: ResolveWriteOperatorId = (ctx, inputOperatorId) =>
+    resolveOperatorIdForWrite(ctx, inputOperatorId, operatorRepo)
 
   // Chain .route() calls so TypeScript infers the full route type tree.
   // hc<AppType> needs this to produce typed client methods.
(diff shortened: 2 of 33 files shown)


jackli921:
Thanks for the review. Addressed the cleanup note in \`d86a5ae\`:

- The unused \`ctx\` in \`VehicleClassService.create\` is pre-existing (introduced by #395, `ff5dd11`, not in this PR's diff), but since it surfaced here I've underscore-prefixed it per biome's suggestion. \`create\` keeps the param for interface symmetry with the other service methods but uses \`SYSTEM_CONTEXT\` for global slug uniqueness, so it's deliberately unscoped. The biome pass of \`bun run lint\` is now warning-free.
- Fair callout on the "biome clean" claim — I'd only run \`biome check\` on the changed files, not the full repo lint. My miss.

The one remaining \`bun run lint\` warning is \`services/booking.ts: 433 lines (cap 400)\` — unrelated, pre-existing, different slice; left out of scope.

PR #426, opened 2026-06-04
test(marketplace): pin OPERATOR_* user-enumeration closures (#396)

### LINKED ISSUE #396: feat(marketplace): operator-scope UserRepository (#386 follow-up)
## Context
Deferred from slice 1 tenancy foundation (#386). `users.operatorId` was added (nullable) in slice 1's migration so that operator staff/owner users can be associated with a tenant. The **UserRepository is not yet operator-scoped**.

Note: unlike vehicles/vehicle_classes, `users.operatorId` is **nullable** — renters and platform admins have no operator. Scoping rules differ accordingly.

## Scope
- Drizzle + in-memory `UserRepository`:
  - Reads: an `OPERATOR_OWNER`/`OPERATOR_STAFF` should only see users within their own operator (`operatorId = ctx.operatorId`). Renters/admins per existing privilege rules. Decide product rule for renters (cross-operator marketplace customers likely remain visible to admins only).
  - Writes that create operator staff: stamp/validate `operatorId`.
- Until then, the repo is covered by `rejectOperatorContextUntilScoped` only where applicable — confirm whether any OPERATOR_* path currently reaches UserRepository and guard if so (slice 1 only guarded booking/message/thread).
- Add cross-operator user isolation integration test.

## Acceptance
- Operator staff cannot enumerate another tenant's users.
- Renter/admin behavior unchanged; unit + integration suites green.

Ref: docs/plans/2026-05-25-marketplace-mvp-proposal.md §5/§6; epic #386.

jackli921:
Closes #396 (merges to `marketplace-pivot`, non-default branch — won't auto-close; close manually on merge).

## What & why

`#396` asked to operator-scope `UserRepository`. Investigation on trunk showed **no `OPERATOR_*` path currently reaches it** — every ingress already blocks operators:

- `/customers*` is `STAFF_ROLES`-gated → operators get **403**
- `/users?ids=` treats non-privileged callers as self-only (`allowedUserIds`) → can't enumerate
- booking paths fail-close at `BookingRepository` (`rejectOperatorContextUntilScoped`) **before** any `userRepo.findByIds`
- `POST /bookings` forces non-staff `renterId = ctx.userId`, so the staff-override `findByIds([renterId])` never runs for a foreign id

Renters are **shared marketplace customers** (`users.operatorId` nullable), so filtering users by `operatorId` would be wrong. Per-operator customer access is deferred to **slice 6**.

So this slice **proves and locks the closures** rather than retrofitting scoping (decision made with maintainer).

## Changes
- `tests/routes/operator-user-isolation.test.ts` (new) — 6 regression locks for the above
- `repositories/types.ts` — documenting comment on `UserRepository` recording the intentional context-free decision

## Verification
- 683/683 API tests pass · full-repo lint clean · `lint:boundaries` / `lint:modules` OK
- `db:verify` N/A — no schema/migration change
- code-reviewer empirically broke each gate and confirmed every test fails for the right reason


jackli921:
Addressed review (both findings):

**P1 — `/users` closure was weaker than documented.** Confirmed: `allowedUserIds` ran the thread lookup with a synthetic `RENTER` context for *all* non-privileged callers, so an `OPERATOR_*` sharing a thread with another user could resolve that user's name. Fixed by returning **self-only for operator roles before the thread lookup** (`routes/users.ts`) — fail-closed until operator messaging is modeled (slice 7). Added a shared-thread regression that was RED before the fix; updated the `UserRepository` doc comment to state the now-true behavior.

**P2 — booking assertion too loose.** Tightened `expect(res.status).not.toBe(201)` → `expect(res.status).toBe(403)` so a future wrong-path failure (e.g. 400) no longer passes while failing to prove the tenant guard.

Commit: `c22a6af`. Verification: 746/746 API tests · full lint clean · boundaries/modules OK.

Issue #60, opened 2026-04-11
bug: edit vehicle dialog doesn't pre-populate pricing fields (#48 regression)


jackli921:
## Symptom

Open the edit dialog on any vehicle in \`/en/manage/vehicles\`. The Daily rate and Hourly rate inputs are empty — they show only the placeholder text — even though the card behind the dialog displays the actual rates (e.g. \`¥6,500/day · ¥900/hr\`).

If the user clicks Save without re-typing the rates, the validator will reject the submission because "at least one rate is required".

## Root cause

\`packages/web/src/components/vehicles/EditVehicleDialog.tsx\` passes an explicit \`defaultValues\` object to \`VehicleForm\` that enumerates every field it wants to pre-populate from the current vehicle. When PR #59 (fleet pricing slice #48) added the \`dailyRateJpy\` and \`hourlyRateJpy\` form fields, it forgot to add them to this list, so the inputs mount with \`undefined\` defaults and appear empty.

\`VehicleForm\` itself correctly wires pre-population from \`defaultValues\` — there's already a test (\`packages/web/tests/components/vehicles/VehicleForm.test.tsx\`) that verifies \`dailyRateJpy: 7500\` in defaults renders in the input. The bug is purely the dialog's whitelist.

## Fix

Add the two new fields to the \`defaultValues\` object in \`EditVehicleDialog\`:

\`\`\`tsx
defaultValues={{
  ...,
  dailyRateJpy: vehicle.dailyRateJpy,
  hourlyRateJpy: vehicle.hourlyRateJpy,
}}
\`\`\`

Add a regression test for \`EditVehicleDialog\` that renders with a vehicle containing both rates and asserts the inputs show the expected values.

## Not fix

- The dialog also fails to pre-populate \`minRentalHours\`, \`maxRentalHours\`, \`advanceBookingHours\`. Those are pre-existing bugs from before #48 — slice #50 (rental rules in form) will touch the same area and is the natural place to fix them together.

jackli921:
Fixed in dd23ec7 (merged to main via #61).

PR #448, opened 2026-06-06
feat(web): pickup-location picker on the vehicle form (#435)

### LINKED ISSUE #435: slice-2 follow-up: expose vehicles.pickupLocationId on the vehicle write path
## Gap

Slice 2 (#387) added the `vehicles.pickupLocationId` DB column + composite FK + index, but never surfaced it through the API. Discovered while building slice 5 (#391): `pickupLocationId` existed *nowhere* in `packages/api/src` or `packages/shared/src` except the schema.

## Done in #391 (step 1, commit 44b2b21)

Read path only: `VehicleBase` type + Drizzle `toVehicle`/`vehicleColumns` projection now carry `pickupLocationId`. Storefront search can read it.

## Still needed (write path)

- `createVehicleSchema` / `updateVehicleSchema` (`packages/shared/src/validators/vehicle.ts`) gain optional `pickupLocationId` (uuid, nullable). Mind the `.partial()`+`.default()` leak (#430/#432 pattern) — no default on the field.
- `routes/vehicles.ts` create/update map it through (currently hardcodes `pickupLocationId: null` with a TODO comment).
- Map FK violation (location belongs to another operator / missing) → 422, like #411.
- Web vehicle edit form: a location picker scoped to the operator's locations.

## Why it matters

Until this lands, the only way to assign a vehicle to a storefront is a direct DB insert (the slice-8 demo seed). Operators cannot place cars at a location through the product, so live storefront search returns empty stores for any operator-created vehicle.

jackli921:
## What

Checkpoint 2 of #435 — exposes `vehicles.pickupLocationId` on the **web** write path. Backend (checkpoint 1, `95a8b17`, in this branch) already accepts `pickupLocationId` on create/update; this adds the UI so an admin can actually place a vehicle at a storefront.

## Changes

- **VehicleForm** — operator-scoped pickup-location `<select>`, a faithful mirror of the existing `classId` picker. The composite FK `(operatorId, pickupLocationId)` means options are scoped to the selected operator when the operator picker shows; the empty option submits `null` (matching the nullable-UUID validator); a stale `pickupLocationId` is cleared on operator change (both the `<select>` onChange and the async operators-sync effect).
- **AddVehicleDialog / EditVehicleDialog** — thread a `locations` prop down; `EditVehicleDialog` pre-selects `vehicle.pickupLocationId` (whitelist guard, same trap as #60/#50/#226).
- **VehicleList** — fetch active locations via `fetchLocationsAction({ includeArchived: false })` (shared `locationKeys.list()` cache with the Locations page) and pass them to both dialogs. Archived storefronts are excluded by the repo's default `list` filter, so they're never offered as targets.
- **i18n** — `form.pickupLocation` + `form.pickupLocationNone` in en/ja/zh.

## Tests (TDD, RED first)

- `VehicleForm.test.tsx` — picker renders None + each location; submits the selected UUID; submits `null` on None; pre-selects in edit mode; scopes options to the selected operator; clears a stale id on operator change.
- `EditVehicleDialog.test.tsx` — pre-selects `vehicle.pickupLocationId`; renders "No pickup location" when null.

## Gate

- web tests **447/447**, tsc ×3, `bun run lint` (exit 0), i18n parity (634 keys ×3), fk-indexes — all green.
- Base `marketplace-pivot` merged in (clean, no overlap); no schema/migration changes, so `db:verify` is N/A.
- code-reviewer: no material findings.

## Follow-up (out of scope, noted)

#444 just fixed `ClassForm`/`LocationForm` to validate edits against the **update** schema; `VehicleForm` still validates edits against `createVehicleSchema`. Same class of issue may apply to vehicles — worth a separate issue.

Closes #435

> Note: base is `marketplace-pivot` (non-default), so `Closes #435` won't auto-fire on merge — close #435 manually after squash-merge.

### DIFF
--- a/packages/api/src/pg-errors.ts
+++ b/packages/api/src/pg-errors.ts
@@ -18,6 +18,15 @@ export const PG_ERROR = {
 export const VEHICLES_CLASS_FK = 'vehicles_operatorId_classId_fk'
 
 /**
+ * Composite FK vehicles(operatorId, pickupLocationId) -> locations(operatorId, id),
+ * named explicitly in schema.ts (#387 slice 2). vehicles carries three FKs, so a
+ * 23503 alone is ambiguous; match on this name to tell a bad (or cross-tenant)
+ * pickupLocationId apart from a bad classId or operatorId (#435, mirrors
+ * VEHICLES_CLASS_FK / #400).
+ */
+export const VEHICLES_PICKUP_LOCATION_FK = 'vehicles_operatorId_pickupLocationId_fk'
+
+/**
  * Composite FK fee_schedules(operatorId, vehicleClassId) -> vehicle_classes(operatorId, id),
  * named explicitly in schema.ts. fee_schedules also carries operatorId -> operators, so a
  * 23503 alone is ambiguous; match on this name to tell a bad (or cross-tenant) vehicleClassId
--- a/packages/api/src/repositories/drizzle/vehicle.ts
+++ b/packages/api/src/repositories/drizzle/vehicle.ts
@@ -95,6 +95,9 @@ export class DrizzleVehicleRepository implements VehicleRepository {
         // classId at the DB (#395).
         operatorId: data.operatorId,
         classId: data.classId,
+        // Storefront placement (#435). Composite FK (operatorId, pickupLocationId)
+        // -> locations seals it to the vehicle's own tenant at the DB (#387).
+        pickupLocationId: data.pickupLocationId,
         name: data.name,
         description: data.description,
         photos: data.photos,
--- a/packages/api/src/routes/vehicles.ts
+++ b/packages/api/src/routes/vehicles.ts
@@ -6,18 +6,31 @@ import {
 } from '@kuruma/shared/validators/vehicle'
 import { Hono } from 'hono'
 import { FLEET_WRITE_ROLES, requireUser, toCallerContext } from '../middleware/auth'
-import { PG_ERROR, VEHICLES_CLASS_FK, pgConstraintName, pgErrorCode } from '../pg-errors'
+import {
+  PG_ERROR,
+  VEHICLES_CLASS_FK,
+  VEHICLES_PICKUP_LOCATION_FK,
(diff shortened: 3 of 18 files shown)


jackli921:
All four findings verified against the code and fixed in `cdd1534`.

- **P1 — admin 400.** Confirmed: `GET /locations` (`routes/locations.ts:43-48`) 400s bypass-scope callers (PLATFORM_ADMIN — in `FLEET_WRITE_ROLES`) without `operatorId`/`includeAll`. The PR mirrored the `classId` call shape, but `GET /vehicle-classes` is a *public* endpoint while locations is scoped — that asymmetry was the bug. Added `includeAll` to `fetchLocations`/`fetchLocationsAction`; `VehicleList` now requests `includeAll: true` (operator-scoped callers ignore it and auto-scope).

- **P1 — archived-location data loss.** Confirmed, and it's a *new* path this PR introduced (edit now always submits `pickupLocationId`). `VehicleForm` now preserves an assigned id that's missing from the active list as a "Current pickup location" option, so a normal save can't silently null it. Covered by a new test that submits unchanged and asserts the id survives.

- **P2 — cross-operator edit options.** Confirmed. `VehicleForm` takes an `operatorId` prop; `EditVehicleDialog` passes the vehicle's immutable `operatorId`, so edit-mode options are tenant-scoped (create mode still scopes by the picked operator).

- **P2 — shared cache key.** Confirmed: `LocationList` uses `locationKeys.list()` with `includeArchived: true`. Moved the picker to its own `locationKeys.assignable()` key to stop archived rows bleeding across the staleTime window.

Tests: `fetchLocations` query-param contract (would have caught P1#1 — the unit tests mocked the action and missed it); `VehicleForm` operator-scoped edit + archived-assignment preservation. Local gate green: web 452/452, tsc x3, lint, i18n-parity (635 x3), fk-indexes.

Note kept out of scope: `ClassForm` has the same cross-operator-edit and archived-assignment shape for `classId`; worth a separate issue rather than widening this PR.

Issue #385, opened 2026-05-27
epic(marketplace): MVP pivot execution plan


jackli921:
# Marketplace MVP pivot execution plan

## Source of truth

Accepted plan: `docs/plans/2026-05-25-marketplace-mvp-proposal.md`

## Goal

Build the marketplace MVP: a multi-tenant Airbnb-style car rental marketplace where partner operators manage storefronts, locations, cars, insurance, pricing, and fees, while renters search storefront-first and book a selected vehicle.

## Locked architecture decisions

- Renter search is storefront-first: search returns storefront/location cards with per-class availability summaries; storefront detail returns available vehicles.
- ACRISS lives on `vehicle_classes.acriss_code`; vehicles reference classes through `vehicles.class_id`.
- Bookings persist `requested_vehicle_id` and `assigned_vehicle_id`; substitutions update assigned vehicle through `VEHICLE_SUBSTITUTED` events.
- Default turnaround is 48 hours after scheduled return; configurable at location/storefront level with vehicle override.
- Booking submit is one DB transaction for availability validation, booking row, initial booking event, and fee snapshot; notifications happen after commit.
- Operator roles never bypass tenant scope. Only `PLATFORM_ADMIN` may bypass scope.
- Operator onboarding is seed plus env-gated admin endpoint for MVP; public self-serve signup is post-MVP.
- Work happens one vertical-slice issue at a time on feature worktree branches from `marketplace-pivot`.

## Slice issues

1. #386 Tenancy foundation
2. #387 Locations
3. #388 ACRISS and vehicle CRUD
4. #389 Insurance, pricing & fees (parent) — #404 4a insurance · #405 4b fees · #406 4c class-pricing removal
5. #391 Renter storefront search
6. #392 Booking and event log
7. #393 Notifications and pre-auth handoff
8. #390 Demo seed polish and E2E

## Preconditions

- Run OpenNext deploy dry-run to decide whether Vite migration slice 1 must precede the MVP demo.
- Schedule Du discovery before slice 4 if possible.
- Create Neon `marketplace-pivot` branch at kickoff.

## Acceptance

- Slice 8 demo shows renter search -> storefront result -> vehicle selection -> booking -> confirmation notification visible in operator portal.
- All slice merge gates in the accepted proposal are green.

jackli921:
[...]

PR #490, opened 2026-06-06
fix(api): disambiguate resend no-ops from real re-sends (#485)

### LINKED ISSUE #485: NotificationService.resend: disambiguate live-lease SENDING from re-sent / already-SENT
Follow-up from slice 7 (#393, PR #482). When `claim` returns null, `processOne` returns the un-claimed row and `resend` reports both 'already SENT' and 'currently SENDING (another sender holds the lease)' as `{ ok: true }` → HTTP 200. Operator sees green with no send. Surface SENDING distinctly so the portal can say 'already in progress' vs 're-sent'. Epic #385.

jackli921:
## Problem
`NotificationDispatcher.processOne` returned the un-claimed row whenever `claim()` declined, so `NotificationService.resend` reported **both** an already-SENT row **and** a live SENDING lease (another sender holds it) as `{ ok: true }` → HTTP 200. The operator saw a green result when **nothing was sent**.

## Fix
`processOne` now returns an explicit `DispatchOutcome`: `sent | failed | already_sent | in_progress | no_recipient`. `claim()` declines **only** a terminal SENT row or a live (non-expired) lease — exactly the `already_sent` / `in_progress` split — so the caller no longer infers "did we send?" from a conflated status. `resend` maps these to a discriminated `ResendResult` carrying an `outcome` the route surfaces (`'resent' | 'already_sent' | 'in_progress'`), so the portal can say "already in progress" / "already sent" instead of a blanket success.

## Tests (TDD, RED→GREEN)
- live SENDING lease → `{ ok: true, status: 'SENDING', outcome: 'in_progress' }`, sender **not** called
- already-SENT row → `{ ok: true, status: 'SENT', outcome: 'already_sent' }`, sender **not** called
- genuine re-send (FAILED→SENT) → `outcome: 'resent'` (updated existing exact-match test)

Gate: api **888** unit · api typecheck · boundaries · biome · web typecheck — all green. No schema change.

Closes #485 (manual close — base is non-default `marketplace-pivot`). Slice-7 follow-up, epic #385.

### DIFF
--- a/packages/api/src/routes/notifications.ts
+++ b/packages/api/src/routes/notifications.ts
@@ -54,6 +54,8 @@ export function createNotificationRoutes(service: NotificationService) {
       // tenant-existence leak. The atomic claim makes a double-click send once.
       const result = await service.resend(toCallerContext(user), c.req.param('id'))
       if (!result.ok) return fail(c, result.error, result.status)
-      return ok(c, { status: result.status })
+      // `outcome` lets the portal distinguish a real re-send from a no-op
+      // ("already in progress" / "already sent") instead of a blanket green (#485).
+      return ok(c, { status: result.status, outcome: result.outcome })
     })
 }
--- a/packages/api/src/services/notification-dispatcher.ts
+++ b/packages/api/src/services/notification-dispatcher.ts
@@ -28,6 +28,20 @@ interface Resolved {
 }
 
 /**
+ * Explicit result of one processOne attempt, so callers (notably the operator
+ * resend) never have to re-derive "did a send actually happen?" from a status
+ * that conflates a fresh send with a no-op. `claim()` returns null only for a
+ * terminal SENT row or a live (non-expired) SENDING lease, which is exactly the
+ * already_sent / in_progress split.
+ */
+export type DispatchOutcome =
+  | { result: 'sent'; row: NotificationLog }
+  | { result: 'failed'; row: NotificationLog }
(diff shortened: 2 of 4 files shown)


jackli921:
## Code review — SHIP

Verdict **SHIP** (no CRITICAL/HIGH). Core correctness verified: `claim()` declines only a terminal SENT row or a live SENDING lease, and that decline predicate is **status-for-status identical** between the in-memory and Drizzle repos — so mapping the un-claimed row's status to `already_sent` vs `in_progress` is sound. Switch is exhaustive (`noImplicitReturns`), `dispatch()`/post-commit caller discard the return (no break), response change is additive for the typed Hono client. Tests assert exact result + `sender.send` not-called (mutation-resistant). Gate: api 888, typecheck/boundaries/biome/web-typecheck green.

### Two LOW notes (cosmetic, not blockers — for the eventual portal/web slice)
1. **TOCTOU:** a row live-SENDING at `upsertQueued` time that another sender `markSent`s before the declined `claim` will read stale `row.status==='SENDING'` → labelled `in_progress` instead of `already_sent` (never the reverse; the send still correctly does not fire). Fix if ever needed: re-read via `findById` after a declined claim and split on that. Not worth the extra query for MVP.
2. **`status` is now derivable from `outcome`** on the success union — the portal should branch on `outcome` (the field this PR exists to add), not `status`.

Issue #458, opened 2026-06-06
feat(marketplace): map + flat-list search results (specific vehicles)


jackli921:
From §1.1 / §5. Add a NicoNico-style **map + left-side list** as a primary search result, coexisting with the storefront-first flow (slice 5).

**Acceptance**
- New renter result: map + list of **specific vehicles** (with plate) across all operators/locations for the search params.
- Storefront-first flow (slice 5) remains; both are first-class ("room to switch").
- Search read-models shaped to support both shapes without rework.
- MVP shows specific vehicles only; class-combo deals are a separate fast-follow issue.

**Design:** reuse slice-5 availability data/API. Reverses proposal §2/§10.12 (flat search was rejected; now first-class).
Refs: epic #385; §1.1, §5.

jackli921:
## Migration renumber needed — trunk is now at 0043

Cross-branch audit heads-up. `origin/marketplace-pivot` now holds:
`0040_vehicle_luggage` (#457) · `0041`/`0042` (#463 fulfillment_mode) · **`0043_add_payment_events` (#461, just merged)**.

Your branch adds **`0041_add_location_coordinates`** (lat/long on `locations`), which collides with trunk's `0041`. (Your `0040` matches trunk's luggage.)

**Renumber target: `0045_add_location_coordinates`.** Recommended merge order is #459 → **#458** → #460, so you land on **0045** (after #459 takes 0044).

Recipe: merge trunk in (no force) → rename the `.sql` + `drizzle/meta/0041_snapshot.json` → `0045_…` → set journal `idx=45` / `tag` / `when` (strictly **> 1781044530577** and > #459's new 0044 `when`) → `bun run db:verify`. Self-contained migration, no SQL-body conflict — number/journal relabel only. (Or regenerate via `db:generate` off merged trunk.)

jackli921:
Merged to marketplace-pivot via PR #513 (squash `8509103`). Slices 0/A/B/C/D/E + both ultrareview P2 fixes (viewport fit-all-pins, accessible row→marker selection). Migration shipped as `0045_add_location_coordinates`.

Issue #412, opened 2026-06-03
slice 6: location archive guard (active bookings) + archived-pickup handling


jackli921:
Follow-up from #387 (slice 2) review (epic #385). Acceptance criterion for slice 6 (bookings).

`LocationService.archive` archives unconditionally (correct for slice 2 — bookings don't reference locations yet). When slice 6 adds `bookings.pickupLocationId`/`dropoffLocationId`:

1. Add an active-booking guard mirroring `VehicleClassService.archive` (count active bookings, 409 with a `code` discriminator, SYSTEM_CONTEXT internal read).
2. Decide handling for vehicles whose `pickupLocationId` points at a now-ARCHIVED location (renter search in slice 5 should not surface them).
3. When the vehicle write path sets `pickupLocationId`, map a cross-tenant/missing-location 23503 to 422 (same as #400 for classId).

jackli921:
Part 1 (active-booking archive guard) merged to marketplace-pivot via PR #491 (squash `55b9a71`):
- `BookingRepository.countActiveForLocation` (pickup OR dropoff, CONFIRMED/ACTIVE) on both repos
- `LocationService.archive` → 409 + `LOCATION_HAS_ACTIVE_BOOKINGS` + `activeBookingsCount`; DELETE route surfaces them

**Remaining (keeping this open):**
- Part 2: hide vehicles whose `pickupLocationId` is ARCHIVED from renter search — deferred to avoid colliding with in-flight search work (#458).
- Part 3: map `23503` (FK violation) → 422 on the vehicle write path when `pickupLocationId` is missing/cross-tenant (mirror #400 for classId).

jackli921:
Starting **part 3** (map cross-tenant/missing `pickupLocationId` FK violation 23503 → 422 on the vehicle write path, mirroring #400's classId mapping). Part 1 shipped in #491. Part 2 (archived-pickup search filter) stays deferred until #458 lands. Branching off `marketplace-pivot`.

jackli921:
**Part 3 done** → PR #492 (test-only; the 23503→422 `pickupLocationId` mapping itself shipped with #435, this adds the missing route-level regression coverage). Verified 7/7 in the `vehicles-route-fk` integration suite against a real Postgres branch.

[...]

Issue #392, opened 2026-05-27
feat(marketplace): slice 6 booking and event log


jackli921:
Parent epic: #385

Source of truth: `docs/plans/2026-05-25-marketplace-mvp-proposal.md`

## Goal

Implement selected-vehicle booking with event log, fee snapshot, turnaround, and substitution-ready schema.

## Scope

- Add append-only `booking_events`.
- Add write-through `bookings.current_state`.
- Replace legacy nullable `bookings.vehicleId` with `requested_vehicle_id` and `assigned_vehicle_id`.
- Enforce exclusion constraint on `assigned_vehicle_id` and effective time range including turnaround.
- Generate `booking_code`.
- Snapshot applicable fee schedules into `bookings.fee_snapshot`.
- Support `VEHICLE_SUBSTITUTED` event path for operator substitutions.

## Acceptance

- Booking submit is one DB transaction for availability check, booking row, initial event, and fee snapshot.
- Concurrent bookings cannot reserve the same assigned vehicle.
- Substitution requires available same-operator, same-location, same-or-better-class vehicle.
- Confirmation page displays booking code, vehicle details, and potential additional charges.

jackli921:
# Slice 6 — Booking & Event Log (issue #392)

**Date:** 2026-06-02
**Status:** Draft v1 — implementation plan for AFK execution; awaiting green light to create worktree
**Parent epic:** #385
**Source of truth:** `docs/plans/2026-05-25-marketplace-mvp-proposal.md` — §6 row 6 (slice scope), §6.1 (E2E gate before slice-6 merge), §6.2 (tenant scoping), §10 item 3 (booking_code format), §9 items 19/22/25, §10 items 9/14/17, §2 rows "Record mutation model" / "Booking write boundary" / "Vehicle substitution" / "Vehicle turnaround buffer".
**Format model:** mirrors `docs/plans/2026-06-02-slice4-insurance-pricing-fees.md` structure/depth, adapted to an event-sourced booking write-path slice.

[...]

PR #491, opened 2026-06-07
fix(api): guard location archive against active bookings (#412 part 1)

jackli921:
## Problem
`LocationService.archive` archived unconditionally — fine in slice 2 (locations weren't on bookings), but since slice 6 (#392/#435) bookings carry `pickupLocationId`/`dropoffLocationId`. Archiving a location still in live use stranded CONFIRMED/ACTIVE bookings pointing at an archived row.

## Fix (mirrors `VehicleClassService.archive`)
- New `BookingRepository.countActiveForLocation(locationId)` — counts BLOCKING (CONFIRMED/ACTIVE) bookings referencing the location as **pickup OR dropoff** (in-memory + drizzle).
- `LocationService` takes `bookingRepo`; `archive()` returns a 409 with a `LOCATION_HAS_ACTIVE_BOOKINGS` code + `activeBookingsCount` when any exist.
- The DELETE route surfaces `code` + `activeBookingsCount` (verbatim mirror of the class route) so the portal can prompt the owner to reassign/cancel first.

## Tests (TDD, RED→GREEN)
- Service: refuses archive with an active pickup booking (409 + code + count); counts a **dropoff-only** reference; **allows** archive when only CANCELLED bookings reference it.
- Repo: `countActiveForLocation` counts pickup OR dropoff, ignores non-active, 0 for unknown.
- Fixed the route-test harness to inject `bookingRepo`.

Gate: api **893** unit · typecheck · boundaries · biome · web typecheck — all green. No schema change.

## Scope
Part 1 (the archive guard) of #412. **Remaining:** part 2 (hide ARCHIVED-pickup vehicles from renter search — deferred to avoid colliding with in-flight #458) and part 3 (map `23503` → 422 on the vehicle write path). #412 stays open for those.

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -444,7 +444,7 @@ export function createApp(overrides?: {
   // inference is retired, so it no longer needs an operator lookup.
   const resolveWriteOperatorId: ResolveWriteOperatorId = (ctx, inputOperatorId) =>
     resolveOperatorIdForWrite(ctx, inputOperatorId)
-  const locationService = new LocationService(locationRepo)
+  const locationService = new LocationService(locationRepo, bookingRepo)
   const insuranceOptionService = new InsuranceOptionService(insuranceOptionRepo)
   const feeScheduleService = new FeeScheduleService(feeScheduleRepo)
   const storefrontSearchService = new StorefrontSearchService(
--- a/packages/api/src/repositories/drizzle/booking.ts
+++ b/packages/api/src/repositories/drizzle/booking.ts
@@ -113,6 +113,22 @@ export class DrizzleBookingRepository implements BookingRepository {
     return row?.value ?? 0
   }
 
+  async countActiveForLocation(locationId: string): Promise<number> {
+    const [row] = await this.db
+      .select({ value: count() })
+      .from(bookings)
+      .where(
+        and(
+          or(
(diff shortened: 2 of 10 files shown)


jackli921:
## Code review — FIX-FIRST, now addressed

Reviewer confirmed the core is sound: the in-memory and Drizzle `countActiveForLocation` count the **identical** set (BLOCKING statuses, pickup OR dropoff), the drizzle `and(or(...), inArray(...))` groups correctly, the route response is additive/non-breaking, and the tests are mutation-resistant (pickup-only, dropoff-only, CANCELLED-ignored).

**One HIGH fixed (`42825ed`):** a missed `new LocationService(repo)` construction site in `tests/integration/locations.test.ts` — the new required `bookingRepo` arg wasn't threaded there. It escaped because api `tsc` only checks `src/**` and the default vitest lane excludes `tests/integration/**`, so it was a latent break in the DB-backed lane. Now injects a real `DrizzleBookingRepository(db)`.

Learn: **Incomplete Fan-Out on Constructor Change** — after changing a signature, grep the symbol repo-wide; don't trust tsc when test globs sit outside `include`.

(LOW: an unrelated pricing test (`bookings.test.ts:959`) is the known clock-skew flake — re-run on red, not caused here.)

Issue #521, opened 2026-06-10
feat(auth): add provider login entry and invite-backed operator access


jackli921:
# Parent

Refines #510. Part of #509 and #385.

# Problem

The Vite app needs separate entry experiences for renters and car rental providers, but there is only one Google OAuth engine underneath. First-time Google users may have no `users` row yet, so the backend cannot safely infer whether they are a renter or an operator from login alone.

Without an invite-backed provider flow, an unknown user could either land in the wrong portal or be allowed to self-select provider access, which is a security and product risk.

# Product Decision

Use separate UI entry points with shared authentication:

- Renter entry: `/[locale]/login`
- Provider entry: `/[locale]/provider/login`
- Provider invite entry: `/[locale]/provider/invite/:token`

Google proves identity. Our database grants authority.

- Unknown renter login creates a `RENTER` user.
- Provider login never creates operator access by itself.
- Provider access requires an existing operator membership or an active provider invite for the verified Google email.
- If provider intent has no matching invite/membership, show a clear "provider access not found" state with sign out and request-access options.

# Scope

- Add provider-specific login UI copy and route in the Vite app.
- Pass login intent, return destination, and optional invite token through the OAuth start/callback flow.
- Add or wire the minimal backend model for provider authorization:
  - `provider_invites` for pre-approved provider emails before first login.
  - `operator_memberships` for accepted/current operator access after first login.
- On Google callback:
  - If intent is renter and no user exists, create `RENTER`.
  - If intent is provider and membership exists, route to `/manage/<operatorSlug>/dashboard`.
  - If intent is provider and invite matches the verified email, create/update user, create membership, mark invite accepted, and route to provider dashboard.
  - If intent is provider and no match exists, do not create provider access.
- Preserve secure `returnTo` handling so redirects cannot leave allowed app origins.
- Add i18n keys for renter/provider login states in `en`, `ja`, and `zh`.

# Non-Goals

- No public provider self-serve onboarding in this issue.
- No full operator portal re-port in this issue.
- No multi-operator role switcher unless it falls out naturally from memberships.
- No separate OAuth provider/client unless required by Google configuration.

# Acceptance Criteria

- Renter login from `/en/login` creates or resumes a renter session and routes to the renter destination.
- Provider login from `/en/provider/login` routes an existing operator member to the correct operator dashboard.
- Provider invite login accepts only the invited Google email and creates the operator membership.
[...]

Issue #512, opened 2026-06-09
feat(web): port operator booking view for marketplace demo handoff


jackli921:
# Parent

Part of #509.

# Problem

The demo runbook expects an operator owner to see bookings after a renter books, but the active Vite business route is still a placeholder. The minimum demo need is not the full legacy business portal; it is a working operator booking view connected to the backend.

# Scope

- Replace the Vite business placeholder needed for the demo with a real operator booking view.
- Connect to the Hono booking/list APIs using the current session and operator scoping.
- Show seeded bookings and newly created renter bookings.
- Include enough date/status/renter/vehicle detail for the operator handoff story.
- Keep tenant isolation visible: operator 2 must not see operator 1 bookings.
- Add i18n keys for `en`, `ja`, and `zh`.

# Acceptance Criteria

- An operator owner can log in and navigate to the Vite operator booking view.
- The view shows bookings for that operator, including a booking created during the demo.
- Cross-operator bookings are not visible.
- Loading, empty, and error states are present.
- Relevant web/API tests or real-DB E2E coverage proves the operator handoff.

jackli921:
Done — merged to marketplace-pivot via PR #537 (squash `682cdb2`). Operator booking list at `/<locale>/manage/bookings` (combined expand=vehicle,renter API + business nav link + tenant-isolation test). Round-1 review P1 (route collision with renter #511) and P3 (queryKey) fixed before merge.

Issue #511, opened 2026-06-09
feat(web): port renter booking create and confirmation flow to Vite


jackli921:
# Parent

Part of #509.

# Problem

The public Vite marketplace surfaces exist for search and storefront detail, but the demo path still needs the authenticated renter booking creation and confirmation screens in the active Vite app.

# Scope

- Add Vite routes for `/[locale]/bookings/new` and `/[locale]/bookings/confirmation`.
- Connect booking creation to the Hono booking API.
- Show selected vehicle/class, pickup/return range, price, insurance options, and potential charges needed for the MVP demo.
- Require renter auth using the existing Vite session guard.
- Preserve the runbook path from search -> storefront -> booking -> confirmation.
- Add i18n keys for `en`, `ja`, and `zh`.

# Acceptance Criteria

- A logged-in renter can open booking creation from a storefront vehicle.
- Booking submit creates a real booking against the seeded Neon branch.
- Confirmation shows booking code and key booking details.
- Unauthenticated users are redirected to login and returned after authentication.
- Relevant web/API tests or real-DB E2E coverage proves the path.

jackli921:
Claiming this — starting work now in a dedicated worktree `feat/511-renter-booking` off `marketplace-pivot`. Builds on the already-merged `_renter` Vite session guard (#497) and the renter booking Hono API. **Other agents: this slice is in-progress, please don't pick it up.** (#510 login is being driven separately via PR #520.)

jackli921:
⚠️ Heads-up from a parallel session: I briefly and mistakenly ran `git reset --hard` against the `feat/511-renter-booking` worktree thinking it was an abandoned 12h-old checkout. I restored it immediately to your last commit, and I can confirm your work is intact and has since advanced to `9f046fc` (wizard→instant-book wiring). No data lost. I've backed off #511 entirely — it's yours. Apologies for the noise.

jackli921:
Done in PR #536 (squash `26bd31d` → marketplace-pivot). Renter instant-book create + confirmation flow (Slices A+B): booking-create/read API client, wizard final-step instant-book submit, confirmation route + view + PreAuthHandoffCard. Closed manually — non-default base means "Closes #511" doesn't auto-fire. Gate green at merge (web vitest 709, e2e, db-drift, test-and-build all pass).

Issue #509, opened 2026-06-09
feat(web): finish Vite UI/backend integration for demo-ready MVP pivot


jackli921:
# Problem

`marketplace-pivot` has the core marketplace backend and public Vite shell running, but it is not fully demo-ready because key authenticated UI surfaces are still incomplete or stranded in the frozen Next.js app. The clearest current blocker is `/en/login`, which renders `Login (port pending)` in the active Vite/TanStack Router app while the demo runbook expects renter and operator login-backed flows.

This creates a mismatch between "MVP pivot features are done" and what a stakeholder can actually click through locally or on a preview: public search/catalog works, but the authenticated demo path is interrupted.

# Execution Split

This issue is the umbrella tracker. Implement as focused child slices:

- #510 — Vite auth/login + session flow
- #511 — Vite renter booking create + confirmation
- #512 — Vite operator booking view for demo handoff

# Goals

- Make the active Vite/TanStack app demo-ready for the marketplace MVP pivot.
- Connect the Vite UI to the Hono API auth/session/backend endpoints instead of relying on frozen Next.js pages.
- Support the demo runbook path: renter search -> storefront -> vehicle -> booking -> confirmation -> operator sees booking.
- Remove or replace placeholder Vite routes that block the demo.
- Add automated coverage so this cannot regress silently.

# Non-goals

- Do not revive the frozen Next.js app as the primary demo surface.
- Do not add new post-MVP product scope such as payments, automated IDV, R2 document retention, or partner revenue reporting unless already required by the MVP demo path.
- Do not change the database schema unless a missing UI integration proves it is required.
- Do not make production DNS/Cloudflare cutover changes in this issue.

# Umbrella Acceptance Criteria

- Visiting `/en/login` no longer shows `Login (port pending)`.
- A seeded or newly created renter can authenticate in the Vite app and remain logged in after navigation.
- The renter can complete the documented marketplace happy path from search to confirmation against a fresh Neon branch seeded with `db:seed` and `db:seed-bookings`.
- An operator owner can authenticate in the Vite app and see the relevant booking in the business/operator UI.
- The demo runbook can be followed from a fresh worktree without knowing hidden branch/env details.
- Local gates pass: `bun run --filter @kuruma/web typecheck`, `bun run lint:i18n-parity`, relevant web/API tests, and the real-DB E2E lane or its documented replacement.

# References

- Epic: #385 marketplace MVP pivot
- Migration epic: #378 Vite/TanStack Router + CF Pages
[...]

PR #537, opened 2026-06-11
feat(web): operator booking list view + combined expand API (#512)

jackli921:
## What

Operator booking list for the business portal (Vite/TanStack-Router shell).

- New route **`/$locale/_business/bookings`** (under the `_business` role guard) rendering `OperatorBookingsView`: a table of **code · vehicle · renter · date range · status · total**, with empty / loading (`PageSkeleton`) / error (retry) states.
- Loader prefetches into the query cache (`ensureQueryData`); the component reads the same options via `useSuspenseQuery` — the locked Vite-shell data pattern.
- Tenant isolation is **server-side** (CallerContext from the session cookie). The web client passes **no** `operatorId`.

## API change: comma-list `expand`

The fulfilling car is only an `assignedVehicleId` UUID on a booking, and `GET /bookings` previously allowed a single `expand` token. This adds a **comma-list** expand:

- `expand=vehicle,renter` enriches each row with **both** projections in one fetch via the new `BookingService.findAllWithVehiclesAndRentersPaginated`.
- Single-token `expand=vehicle` / `expand=renter` is unchanged (the route now parses `expand` into a trimmed `Set`).
- The combined method reuses `findAllPaginated`, so it inherits the exact same three-way tenant scope — a foreign tenant's booking is dropped **before** either enrichment runs.

## Tests (TDD)

- `api/tests/routes/bookings.test.ts` — combined `expand=vehicle,renter` returns both projections.
- `api/tests/routes/operator-user-isolation.test.ts` — **tenant-isolation lock** for the combined path: an operator sees only its own booking, both projections resolve from its own tenant, and the foreign renter is never looked up (`SpyUserRepository`). Mirrors the existing `expand=renter` isolation test (#396).
- `web/tests/vite/operator-bookings/api.test.ts` — client builds `expand=vehicle,renter`, maps rows (incl. `vehicleName`), null-vehicle/null-renter fallbacks, `operatorBookingsQueryOptions`.
- `web/tests/vite/operator-bookings/OperatorBookingsView.test.tsx` — rows, JST range, email fallback, em-dash for missing data, per-status label, empty state.
- i18n `bookings.operator.*` added for **en / ja / zh** (parity green).

## Gate (local, all green)

API 1076 · web 693 · lint 0 · typecheck 0 (web default+app, api) · boundaries 0 · fk-indexes 40/40 · i18n-parity 737×3 · build 0 · dist-size 12.7% of budget. Code-reviewer: **SHIP**, no CRITICAL/HIGH.

## Notes / follow-ups

- **Tenant-isolation verification** is done as a full-stack `createApp` integration test (real JWT + repos) rather than a browser `test:e2e:real-db`, because operator login is not yet wired into the Vite shell (that's #521, a separate worktree) — there is no way to authenticate an operator in the browser harness yet. The integration test exercises the same end-to-end scoping guarantee. A browser e2e can follow once operator login lands.
- `operatorBookingsQueryOptions` keys by `status` only (not `limit`); latent today since no caller passes `limit`. Tighten the key when filter UI is wired.
- Potential future path collision: a renter "my bookings" list, if added at `/$locale/bookings`, would clash with this `_business/bookings`. No such renter route exists today (#511 covers create+confirm, not a list).

Refs #512 #509


jackli921:
Thanks — both addressed in `2e9e93d` (after merging trunk to pick up #511).

**P1 (route collision → operator view unreachable).** Confirmed: #511 merged `_renter/bookings.tsx` to trunk, which shares the `/$locale/bookings` URL with my pathless `_business/bookings.tsx`. Fixed by moving the operator view to **`/$locale/manage/bookings`** (`_business/manage/bookings.tsx`) — the `manage/` segment matches the frozen app's business routes and the documented business-route convention. Regenerated `routeTree.gen.ts` now maps `/$locale/bookings` → renter and `/$locale/manage/bookings` → operator (distinct entries). Also **wired it into the business navbar + mobile menu** so an operator owner can actually navigate to it, and strengthened `Navbar.test.tsx` to assert the `/$locale/manage/bookings` link is present for a business user.

**P3 (queryKey omits `limit`).** `operatorBookingsQueryOptions` now keys on `['operator-bookings', status, limit]`; test updated.

Gate green after the merge: web 722 · api 1076 · lint 0 · typecheck 0 (web default+app, api) · boundaries 0 · i18n-parity 742×3 · build 0 · dist-size 12.8%.

PR #538, opened 2026-06-11
fix(web): renter search defaults + date retention (search UX)

jackli921:
## Summary
Fixes two renter-search UX bugs reported on the Vite shell.

### Bug 1 — search → 500 (environment fix, no code)
The dev DB was 5 migrations behind (43 of 48), missing `0045_add_location_coordinates`. `/storefronts/search` SELECTs `locations.latitude`, which didn't exist → Postgres rejected the query → 500. It only surfaced *with* dates because the loader only calls the endpoint when a valid range is present. Resolved by running `bun run db:migrate` (db:verify 4/4, endpoint returns 200). **No code change.**

### Bug 2 — search defaults + date retention (TDD, 3 commits)
Locked behaviour:
- Pickup defaults to the **next whole JST hour**, return to **+3 days**.
- `/search` **auto-runs** with the seeded range (no blank initial state).
- Dates are **retained on back-navigation** — search page via the URL range, hero via `sessionStorage`.

Commits:
- `289f841` — pure helpers: `formatJstDateTimeLocal` (`lib/datetime.ts`), `defaultSearchRange(now)` + `searchRangeToSeed(search, now)` (`vite/storefronts/params.ts`), `now` injected. +10 tests.
- `10f571b` — hero `SearchWidget` prefills `readPersistedRange() ?? defaultSearchRange()` and persists on submit; new `vite/storefronts/storage.ts` (sessionStorage). +9 tests.
- `762c1c0` — `$locale/search.tsx` `beforeLoad` seeds defaults + redirects (replace) when a bound is absent → auto-run; `StorefrontSearchForm` keyed on `${from}|${to}` so back/forward remounts the inputs (no stale/cleared values); search form persists on submit. +1 test.

## Test plan
- [x] Full web suite — `bun run test` → **699 passed (116 files)**
- [x] Root lint — `bun run lint` → biome clean, module boundaries pass
- [x] Vite build — `bun run --filter @kuruma/web build` → exit 0; `routeTree.gen.ts` unchanged
- [ ] Manual (dev servers on :3001/:8787):
  - Home hero prefilled (next hour / +3d); search → results; browser back → hero keeps dates.
  - `/search` with no params → auto-redirects to seeded range and shows results.
  - Change dates on `/search`, submit, browser back → previous dates restored.

## Notes
- Targets **marketplace-pivot**, not `main` (pre-pivot prod). No "Closes #" — user-reported ad-hoc batch, no issue.
- Optional follow-up (not done): `packages/api/src/error-handlers.ts:25` logs only `err.message`, swallowing `err.cause` — logging the cause would make DB errors like Bug 1 self-diagnosing.


jackli921:
## Hardening pass (code-reviewer agent)

Verdict: **ship-ready** — no CRITICAL/HIGH. All four focus areas (JST timezone math, redirect-loop convergence, sessionStorage safety, validation) verified correct, incl. empirically under a non-JST host clock.

**Applied (commit e4566d4, pure refactors, suite stays 699/699):**
- Dedupe hero `initialRange()` — was reading sessionStorage + building a Date twice on mount.
- Name the bare `24` as `HOURS_PER_DAY`.

**Conscious decision — NOT changed:** when a renter arrives with only one bound (e.g. `?from=X` and no `to`, from a hand-edited/stale link), the seed replaces it with the full default range rather than preserving the present bound. Preserving-and-deriving the missing bound adds branches + a parse/add/format helper + malformed-input fallback for a rare case; the current behavior is always a valid positive range. Keeping it simple (YAGNI).

**Out-of-scope follow-up (not in this web PR):** `packages/api/src/error-handlers.ts:25` logs only `err.message`, swallowing `err.cause` — that's why Bug 1's original 500 log was an unhelpful truncated SQL dump. Worth a separate API issue.

Issue #551, opened 2026-06-12
feat(marketplace): turnaround 60min floor + realistic seed + renter cooldown surfacing


jackli921:
## Background — this is **not** a scheduling-logic bug

Reported symptom: after a renter books a car (e.g. Suzuki Wagon R), re-searching makes that car — and eventually the whole store card — disappear, as if a booking is single-use.

Root-cause investigation (marketplace-pivot) showed the scheduler is **correct**:
- Overlap predicate is a proper range-overlap, scoped to the requested window: `NOT EXISTS (booking WHERE status IN (CONFIRMED,ACTIVE) AND tstzrange(startAt, effectiveEndAt) && tstzrange(from,to))` (`repositories/drizzle/availability.ts`).
- Booking does **not** flip `vehicles.status` — it stays `AVAILABLE`.
- `effectiveEndAt` is `NOT NULL`, set by the trigger (migration `0037`) to `endAt + COALESCE(turnaround, 2880) * interval '1 minute'` — bounded, correct unit math.
- The gist exclusion constraint `bookings_no_overlap` already guarantees zero overlap at **any** turnaround.
- CI-green integration test `availability.test.ts › 'includes vehicles without overlap'` proves a non-overlapping later date returns the car.

The real cause of the illusion: **every seeded location uses the 48h (2880 min) default turnaround**, and the renter search always defaults to / restores a near-future window. A booking blocks the car for `rental + 48h` (~5 days), so every near-future re-search keeps overlapping it. The car *does* return for any pickup ≥ `endAt + 48h`.

## Decision (source of truth: `docs/plans/2026-05-25-marketplace-mvp-proposal.md` §9 item 20, §10 item 11, §2)

The proposal already decided turnaround is **operator-adjustable per location** (default 48h), with an optional per-vehicle override for exceptions. It is *not* universal — `locations.defaultTurnaroundMinutes` is per-row and already editable in the operator `LocationForm`. The seed merely sets them all to 48h.

Choices confirmed for this slice:
1. **Keep 48h as the documented default, but vary the seed** with realistic per-location values so availability looks normal.
2. **Enforce a 60-minute minimum floor** (validator + DB CHECK). A floor does **not** add overlap-safety (already guaranteed) — it guarantees a servicing gap (clean/refuel/handoff) and stops an operator misconfiguring 0.
3. **Defer the per-vehicle override** (docs: MVP = location-only, migration `0037`). Revisit as a separate issue if a real exception appears.
[...]

Issue #531, opened 2026-06-10
feat(operator): geocoding foundation — provider-neutral Geocoder port + capture location lat/lng on save


jackli921:
## Context

`locations` already has `address` (notNull), `latitude`/`longitude` (`doublePrecision`, **nullable**, migration `0045` / #458), `operatingHours`, and `defaultTurnaroundMinutes`. The search **read path is already wired**: `flat-search.ts` `toResultLocation` maps lat/lng into `ResultLocation`, and `SearchMapList.geocodedByLocation` plots geocoded rows and drops null-coord rows (the list still shows them). Map rendering is behind `packages/web/src/vite/search/MapAdapter.ts`.

The gap is purely the **write/capture + contract** side: `routes/locations.ts` hardcodes `latitude:null/longitude:null` ("not captured yet"), and `createLocationSchema`/`updateLocationSchema` don't accept coords. Every operator-created location is null-coord → filtered off the map. Only **seeded** points appear today.

## Scope split (resolves the #529 contradiction)

- **#531 (this) = API / service / contract ONLY — no UI.**
- **#529 = operator location-form UI** (address input + manual pin / paste-coords override) that calls this contract. **#529 depends on #531.**

Every acceptance criterion below is closeable at the service/contract level without UI.

## Provider-independence (the decision this locks)

1. **lat/lng (+ the operator-submitted `address`) is the canonical source of truth.** `address` is authoritative and **never overwritten** by geocoder output. **Never key any entity off a provider `place_id`.**
2. **`Geocoder` port** — `geocode(address: string) -> { lat: number; lng: number } | null`. (Trimmed: no `formattedAddress`/`precision` for MVP — address is authoritative and we persist no precision; extend the port if precision-gating is needed later.) Business logic depends on the interface; the concrete adapter is wired **only in `index.ts`** (same DI as repositories + `MapAdapter`). Provider swap = one line there (proven by a fake-`Geocoder` DI test).
3. **`NominatimGeocoder` (OSM) first — forward geocoding only.**

## Provider constraints — OSMF Nominatim policy (MUST encode)

Public Nominatim ([OSMF policy](https://operations.osmfoundation.org/policies/nominatim/)): **≤ 1 req/s**, descriptive **User-Agent** (app id + contact), **attribution**, **caching**, and **forbids autocomplete**. Geocode-on-save complies (low volume, one call per save, coords persisted = cached). The adapter sets a proper User-Agent; no bursts.

[...]

Issue #523, opened 2026-06-10
epic(web): operator/provider portal — Vite re-port


jackli921:
# Why

The web app is mid-migration from Next.js to Vite (#378). Only the **renter marketplace** was ported. The **operator/provider portal was never ported** — it exists only in the frozen Next.js tree (`packages/web/src/app/[locale]/(business)/manage/*`) and is not served by the Vite app. A logged-in provider currently lands on a placeholder `Dashboard (port pending)` stub, so a provider cannot actually run their business. This epic tracks porting the full operator portal to the Vite shell.

The Hono API already serves this data — most of this is a UI re-port, **not** new backend product work. (One exception: operator **locations management UI is net-new** — see that child.)

# Relationship to other work

- Part of #385 (marketplace MVP) and #378 (Vite migration).
- **Login/auth is decided in #521** (separate provider entry + invite/membership authorization). Provider login routes to `/manage/<operatorSlug>/dashboard` — that destination only becomes useful once this epic lands. #521 also moves operator authorization to `operator_memberships`; the `_business` route guard in every child below must gate on **membership, not `users.role`**.
- #509 stays the thin demo bridge (login + renter booking + operator *sees* a booking, #512). This epic is the full provider product surface and deliberately does **not** extend #509.

# Child issues

- [ ] #524 operator dashboard (overview stats)
- [ ] #525 operator bookings — calendar + detail (renter info folded in)
- [ ] #526 fleet — vehicles list + CRUD + photos
- [ ] #527 vehicle detail — calendar-lite + utilization
- [ ] #528 vehicle classes — CRUD
- [ ] #529 locations/storefronts — management (net-new UI)
- [ ] #530 pricing config — insurance + fees + per-vehicle

# Out of scope (MVP — decided)

- **Customers / CRM screen** — cut. Renter info surfaces inside **booking detail only** (name, contact, booking history if cheap). No standalone customer-management screen.
- **Chats / messaging** — cut. Needs persistence, unread counts, RLS/tenant safety, and notification behavior; backend #300 stays **parked**. Operator contact for MVP = booking detail + email/notification surfaces. Post-MVP.
- Public provider self-serve onboarding — invite-only for MVP, per #521.

# Source of truth

[...]

Issue #445, opened 2026-06-06
ci: gate marketplace-pivot on the authenticated real-DB e2e track (Neon branch lifecycle)


jackli921:
**Slice 2 of #416.** Slice 1 (PR linked below) landed the authenticated real-DB e2e harness (`playwright.real-db.config.ts` + `e2e/real-db/`), verified locally. This issue is the remaining work: run it in CI as an enforced gate.

## Blocked on
- **`NEON_API_KEY` is not in GitHub Secrets** (only `AUTH_SECRET` + `PROD_DATABASE_URL` exist). A Neon API key with branch create/delete scope is required to spin a disposable branch per CI run. Add it before this can go green.

## Scope
1. New `e2e-real-db` job in `.github/workflows/ci.yml`, separate from the mock `e2e` job.
2. Create a disposable Neon child branch off `marketplace-pivot` (neonctl or the Neon API), capture its connection string.
3. `db:migrate` + `db:verify` + `db:seed` against it.
4. `AUTH_SECRET=${{ secrets.AUTH_SECRET }} DATABASE_URL=<branch> bun run test:e2e:real-db`.
5. Delete the branch in an `always()` teardown step.
6. Add the job to branch protection (required check) on `marketplace-pivot`, alongside `test-and-build` + `db-drift`.

## Notes
- The harness already throws if `AUTH_SECRET`/`DATABASE_URL` are absent and uses dedicated ports (web 3002 / api 8788), so it slots into a CI job without touching the mock track.
- Coverage to extend once green: operator write-scope isolation (#397/#402) and the slice-8 demo flows (#390).

jackli921:
Promoted to a **prerequisite sub-task of #390** (interim core-path integration), not just a P2 nice-to-have. The revised slice-8 plan (`docs/plans/2026-06-02-slice8-demo-seed-e2e.md` §5.2) makes the **real-DB** happy-path the required §6.1 merge gate, which needs this lane: `playwright.real-db.config.ts` (real API+web → seeded ephemeral Neon branch), session helper, migrate→seed→test→drop lifecycle. The existing mock-API specs stay as a fast smoke check.

jackli921:
## Implementation plan — Slice 2 of #416 (revises the original scope)

**Decision (with @li.jack0707): use a local `postgres:16` service container, _not_ a Neon branch.**

### Why the pivot from the Neon-branch scope above

[...]

PR #575, opened 2026-06-12
feat(operator): locations/storefronts management (Vite port) — #529

### LINKED ISSUE #529: feat(web): build operator locations/storefronts management in Vite
Part of #523. Part of #385, #378. Auth model per #521.

# Goal

Operator locations/storefronts management: list / add / edit / archive pickup locations (incl. operating hours and turnaround/buffer).

# IMPORTANT — net-new UI

Unlike the other children, there is **no Next.js source page** for a standalone locations admin (verified: no `manage/locations` route and no Location/Storefront form components). Pickup locations are currently only set via a picker inside vehicle/class forms (#407 / #435). The **backend exists** (slice 2 #387), so this is **new UI on an existing API** — scope accordingly (no straight port).

# Scope

- Locations list (name, address/region, operating hours, status).
- Add / edit / archive location (default 48h turnaround per proposal §4).
- Archive guard: block/handle archiving a location that has active bookings (#412).

# Acceptance criteria

- Operator can list, add, edit, and archive locations; persists via API.
- Archiving a location with active bookings is guarded (#412).
- `_business` guard (membership per #521) redirects non-members.
- Tests cover list, create, edit, archive-guard, and route guard.

# Related

#387 (locations backend), #435 (pickup-location write path), #412 (archive guard), #394 (region taxonomy — post-MVP).

jackli921:
Closes #529. Part of the #523 operator-portal epic.

## What
Ports the frozen Next.js `modules/locations` admin to the live Vite shell as full CRUD for operator pickup/return locations, at `/<locale>/manage/locations` (behind the `_business` guard).

> **Scope note:** the issue states "no Next.js source, net-new UI" — that's incorrect. A complete, tested Next.js locations admin already exists from #387. This PR is a **port**, not an invention, so it reuses the existing `business.locations.*` i18n and mirrors the field/validation contracts.

## Slices
- **1 — Read-only list:** operator-scoped list + status badges + empty state (already on branch).
- **2 — Create:** "Add location" → `AddLocationDialog` (POST, inline 409 duplicate-name).
- **3 — Edit:** row Edit → `EditLocationDialog` (PATCH, prefilled from the row, no GET round-trip).
- **4 — Archive:** row Archive → `ArchiveLocationDialog` (DELETE soft-archive; surfaces the `LOCATION_HAS_ACTIVE_BOOKINGS` 409 with a localized active-booking count; button disabled for already-archived rows).

## Design
- **Cookie-based + operator-scoped.** The client never sends `operatorId`; tenant scoping is server-side via the session (`CallerContext`), so a cross-tenant write is impossible by construction. Mirrors `operator-fleet`.
- **FC/IS.** `OperatorLocationsView` stays a pure projection of the rows + row-action callbacks; the route owns all dialog state and the loader/suspense boundaries.
- **Coordinates out of scope** (geocoding is #531) — the create/update validators on trunk carry no lat/lng.
- **Archive count surfacing.** The generic `unwrap` drops the `activeBookingsCount` discriminator, so `archiveLocation` reads the body directly and throws a typed `LocationArchiveBlockedError` carrying the count; the dialog renders it via a new ICU-plural key `business.locations.archiveBlocked` (added to en/ja/zh).
- Post-rebase: aligned the turnaround input floor to `min={60}` to match #551's 60-minute Zod/DB floor.

## Backend
Zero changes — all CRUD + the archive guard already live on `marketplace-pivot`.

## Tests / gates (local)
- Web suite: **842 passed** (17 new operator-locations tests: list/create/edit/archive + row-action wiring).
- `tsc --noEmit`: clean (web + api). Biome: clean. Module boundaries + file-size: clean.
- i18n parity: **812 keys × 3 locales** (balanced; +1 `archiveBlocked`).
- Route tree regenerated via `vite build` (not `tsr generate`).

## Not covered
- Live browser smoke of the dialogs (manual).
- Real-DB E2E for `/manage/locations` (tracked as a #445 follow-up).


jackli921:
### Review: action-surface authz parity (same design class as #525)

The **read** path here is handled well — `fetchOperatorLocations` uses `includeAll=true` precisely so `GET /locations` doesn't 400 for bypass-scope roles, and never names an `operatorId` so an operator's read stays tenant-scoped. 👍

The gap is on the **write/action** side, and it mirrors a finding we just captured on #525:

> `_business` only admits a user to the business *shell*. Each CRUD/action surface still has to match the backend's actual authorization + scoping contract. In #525 that means **don't show "New Booking" to roles that would 403** (and would reopen the #475 user-enumeration vector). Same shape here: **don't show "Add Location" to `STAFF/ADMIN/PLATFORM_ADMIN` unless the UI supplies the required `operatorId`.**

Concretely on this branch:
- `OperatorLocationsView` renders the Add-Location surface with **no role gate** (no `role`/`isOperator`/`session` check in any `vite/operator-locations/*` file).
- `createLocation` → `POST /locations` sends `CreateLocationInput` with **no `operatorId`** and relies on session-tenant inference ("the session cookie scopes the write"). That's correct for `OPERATOR_*`, but a bypass-scope role has no tenant to infer, so their create can't succeed.

So a `STAFF/ADMIN/PLATFORM_ADMIN` user is shown an action the backend can't fulfill for them. Please fix rather than leaning on "the backend rejects it" — two acceptable directions:

1. **Make locations `OPERATOR_*`-only on the surface** — gate the Add/Edit/Archive actions (and ideally the nav entry) by role so bypass roles get the read-only view, or
2. **Support bypass roles properly** — add an operator selector/default and include `operatorId` on create (confirm `POST /locations` + the `CreateLocationInput` validator accept an explicit `operatorId` for bypass scope).

Either is fine; #1 is the smaller change if operator self-service is the only intended writer for now. Flagging so it's decided explicitly, not left as an implicit 403.

Issue #304, opened 2026-04-17
chore: Cloudflare account migration — personal → freelance


jackli921:
Transfer Workers from personal CF account to freelance CF account.

**Runbook:** `docs/cloudflare-account-migration.md`

## Steps
1. Create new CF account on freelance Gmail (Free plan)
2. Generate API token (Edit Cloudflare Workers template), copy Account ID
3. Replace `CLOUDFLARE_API_TOKEN` + `CLOUDFLARE_ACCOUNT_ID` in GitHub Secrets
4. Run `gh workflow run deploy.yml` — creates both workers on new account
5. Verify `/health` on API, `/en` on web, login + fleet page
6. Update `WEB_ORIGIN` (wrangler.toml) and `API_WORKER_URL`/`WEB_WORKER_URL` (deploy.yml) to new subdomain
7. (Optional) custom domain setup
8. Delete workers from old account, revoke old API token

App secrets (\`AUTH_SECRET\`, \`PROD_DATABASE_URL\`, \`AUTH_GOOGLE_*\`) stay the same.

Needs human to do the CF dashboard signup + token generation.

jackli921:
Parked per owner decision (2026-06-16): the CF account migration to a freelance account is **not happening** — the beta ships on the existing personal CF account (sole-proprietor). Removed from the Pre-launch milestone. Reopen/re-milestone only if a future business-structure change forces account consolidation.

jackli921:
Closing as won't-do. The account migration personal→freelance is not happening (per 2026-06-20 decision; already removed from milestone).

The two issues gated on #304 — #601 (prod KV flip) and #500 (CSP enforce) — are not orphaned: they now hang off **#1009**, which provisions the needed CF resources (KV namespace + live preview) on the **current** account.

Reopen/re-milestone only if a future business-structure change forces the migration.

PR #591, opened 2026-06-12
feat(api): app-side global throttle for forward geocoding (#574)

### LINKED ISSUE #574: feat(api): app-side rate-limit + cache for forward geocoding before enabling Nominatim in production (#531 follow-up)
Follow-up to #531 (geocoding foundation, PR #564). Surfaced in review of #564.

## Problem
`NominatimGeocoder` makes a real upstream call inline on every address-only location create, address change, or `regeocode:true`. The public OSM/Nominatim [usage policy](https://operations.osmfoundation.org/policies/nominatim/) sets an **absolute maximum of 1 request/second across the whole application**, forbids bursts/autocomplete, and asks consumers to **cache** results. There is currently **no app-side throttle or cache** — concurrent operator saves can breach 1 req/s immediately and get the app's IP rate-limited or banned.

CF Workers run as many isolates, so an in-process token bucket cannot enforce a *global* rate — a correct limiter needs shared state (Durable Object or KV).

## Current mitigation (shipped in #564)
- Geocoder is **disabled by default**: a real provider is constructed only when BOTH `NOMINATIM_USER_AGENT` and an explicit `NOMINATIM_API_URL` are set (no silent default to public OSM). Otherwise a null-returning stub runs (locations save with no coords — best-effort).
- `.env.example` + the `index.ts` factory comment warn that public OSM is local/demo-only and production must use a self-hosted/proxied/quota-backed endpoint.
- Non-OK responses (429/403) and exceptions now `console.warn` so throttling/bans are observable.
- Coords are persisted and only re-derived on address-change / explicit `regeocode`, so steady-state re-saves don't re-hit the provider (a coarse cache).

## Scope (this issue) — required before enabling live geocoding at scale
- A **global rate limiter** (DO- or KV-backed token bucket, ~1 req/s) wrapping the `Geocoder` port, OR
- Wire production to a **quota-backed provider** (Google/Mapbox adapter — the port already supports a one-line swap), and/or a **self-hosted Nominatim**.
- A **forward-geocode response cache** (KV, keyed by normalized address) to satisfy the policy's caching ask and cut duplicate lookups across locations.
- Decide + document the production provider choice.

## Blocking relationship
This should land **before** #529 (operator location-form UI) drives real geocoding volume in production. Until then the null stub keeps prod safe.

jackli921:
## What & why

Before we can point the operator-location geocoder (#531) at the public OSM/Nominatim endpoint, we need an app-side guard: OSMF's usage policy caps the **whole app** at 1 req/s and forbids bursts. A burst of operator saves (or a seed/import run) would breach that and earn an IP ban. This PR adds that guard.

**Approach — decorator on the port (DIP + Decorator):** a `ThrottledGeocoder` wraps the existing `Geocoder` port. It checks a Cloudflare-native `[[ratelimits]]` binding (`GEOCODE_LIMITER`, 1 req / 10s) keyed by a single constant (`geocode:global`) so the cap is **global**, not per-IP. Over the limit ⇒ skip the lookup and return `null`. A skip is byte-identical to a provider miss, which `LocationService` already handles — the location still saves, just with no map pin (a later edit/`regeocode` fills it). The CF-specific `limit({ key })` shape is adapted to the port's `limit(key)` at exactly one line in `index.ts`, so nothing in `services/` knows about Cloudflare.

**Fail-OPEN:** if the limiter *itself* errors, geocoding proceeds unthrottled — a rate-limit-service blip shouldn't kill all geocoding. This is safe **only while public OSM stays opt-in** (gated on `NOMINATIM_USER_AGENT` + `NOMINATIM_API_URL`); production runs LocationIQ, whose own server-side cap is the real guarantee. The throttle is best-effort defense-in-depth.

## Scope note

#574's title covers "rate-limit **+ cache**". This PR delivers the **rate-limit** half (the production blocker). The **KV cache** half is split into a follow-up (gated on the #304 account migration that also unblocks R2/KV), filed on merge.

## Changes
- `services/geocoding/throttled-geocoder.ts` (new) — the decorator + fail-open/never-throw logic
- `services/geocoding/types.ts` — new `RateLimiter` port (`limit(key) → { success }`)
- `services/geocoding/nominatim-geocoder.ts` — optional `apiKey` → `?key=`, making LocationIQ a pure env-var swap
- `index.ts` — DI wiring + CF-binding adapter (the only Cloudflare-aware line; still 797/800 cap)
- `wrangler.toml` — `GEOCODE_LIMITER` binding (ns 1005, 1/10s)
- `.env.example` — demo (OSM) vs prod (LocationIQ) config + dev-unthrottled warning

## Reviews (run on the code, not just the plan)
- **code-reviewer → SHIP.** 0 CRITICAL / HIGH / MEDIUM. 3 LOW, all acceptable (JSON-escaped address log; `globalThis` cast consistent with 4 sibling limiters; LocationIQ requires `?key=` over HTTPS, never logged).
- **architect → SHIP.** No fix-now blockers. Two ~1-line doc hardenings folded in (`3bdb2b1`): fail-open's OSM-opt-in dependency, and the limiter being per-colo not cross-region.

## Known residuals
- **R1** — CF fixed-window can admit 2 req across a window boundary. Accepted: immaterial for dozens of lifetime lookups behind a provider's own cap.
- **R2 / F6** — bulk onboarding silently drops pins (only a `console.warn`, no durable marker). Deferred to the cache/retry follow-up. Scope note carried to that issue: the fix needs a **distinguishable pending state** (`null` today conflates "throttled, retry works" with "un-geocodable, retry won't") — not a blind retry-everything sweep.
- **R3** — `globalThis.GEOCODE_LIMITER` resolution isn't unit-proven, but it's the **5th instance** of a pattern 4 sibling limiters already run in production (`nodejs_compat` hoists bindings onto `globalThis`). Covered by the pre-prod smoke below.

## Pre-prod smoke (manual, before enabling in prod)
1. `wrangler dev` → confirm `typeof globalThis.GEOCODE_LIMITER !== 'undefined'`.
2. POST two locations <10s apart with a real `NOMINATIM_*` config → assert the **2nd** persists `latitude: null` (proves the binding throttles the real path, not just that it exists).

## Test plan
- `bun run --filter @kuruma/api test` → **1209 pass** (108 files), incl. 6 new `ThrottledGeocoder` unit tests (allow/skip/fail-open/never-throw/shared-key), 2 `NominatimGeocoder` key-param tests, 1 DI test proving an over-limit binding skips the lookup while the save still 201s with null coords.
- `tsc --noEmit` clean · `lint:boundaries` OK · biome clean.

Closes #574

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,11 +12,20 @@
   # real admin address(es), not a placeholder.
   PLATFORM_ADMIN_EMAILS=you@yourdomain.example
 
-  # Forward geocoding for operator locations (#531). DISABLED unless BOTH are set
-  # (otherwise the API persists locations with null coords — safe, just no map pin).
-  # WARNING: the public OSM endpoint (https://nominatim.openstreetmap.org) caps the
-  # WHOLE app at 1 req/s and forbids bursts — there is NO app-side throttle/cache yet
-  # (#574). Use it only for local/demo; in production point NOMINATIM_API_URL at a
-  # self-hosted / proxied / quota-backed instance and add a limiter first.
+  # Forward geocoding for operator locations (#531). DISABLED unless BOTH a
+  # User-Agent AND an endpoint are set (otherwise locations save with null coords —
+  # safe, just no map pin). Outbound lookups are throttled app-wide by the
+  # GEOCODE_LIMITER binding (#574, ~1 req/10s) so a burst can't breach OSMF's 1 req/s.
+  #
+  # Demo / low volume — public OSM (free, no account; throttle keeps it policy-safe):
   # NOMINATIM_USER_AGENT="kuruma-rental/1.0 (ops@yourdomain.example)"
   # NOMINATIM_API_URL=https://nominatim.openstreetmap.org
+  #
+  # Production — LocationIQ (Nominatim-compatible drop-in, free 5k/day, storage OK):
+  # keep the UA, point the URL below, add a key. A real provider's own server-side
+  # limit is the actual guarantee; the app throttle is best-effort defense-in-depth.
+  # NOMINATIM_API_URL=https://us1.locationiq.com/v1
+  # NOMINATIM_API_KEY=your-locationiq-token
+  #
+  # NOTE: local dev has no GEOCODE_LIMITER binding (unthrottled). Point at LocationIQ
+  # or a self-host before any bulk seed/import so you don't trip an OSM IP ban.
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -141,6 +141,7 @@ import { FeeScheduleService } from './services/fee-schedule'
 import { FlatSearchService } from './services/flat-search'
 import { FleetOverviewService } from './services/fleet-overview'
 import { NominatimGeocoder } from './services/geocoding/nominatim-geocoder'
+import { ThrottledGeocoder } from './services/geocoding/throttled-geocoder'
 import type { Geocoder } from './services/geocoding/types'
 import { GoogleTranslationProvider } from './services/google-translation-provider'
 import { InsuranceOptionService } from './services/insurance-option'
@@ -199,6 +200,9 @@ export function createApp(overrides?: {
   photoUploadLimiter?: RateLimitBinding
   photoUploadUserLimiter?: RateLimitBinding
   publicCatalogLimiter?: RateLimitBinding
+  // Over-limit ⇒ the geocoder skips the lookup (#574). Inject a deny-binding in
+  // tests; absent ⇒ the globalThis-resolved GEOCODE_LIMITER (or unthrottled dev).
+  geocodeLimiter?: RateLimitBinding
   // Injected Google OAuth runtime (provider + account store). Integration tests
   // pass a fake so the callback can be exercised without a live Google/DB.
   googleAuthRuntime?: GoogleAuthRuntime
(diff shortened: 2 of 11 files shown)


jackli921:
## Up-to-date re-review (post trunk-merge)

Brought the branch current with `marketplace-pivot` (merged in the web-only #587, no conflict) and ran an **independent adversarial second pass** over the up-to-date diff. It tried to break the first pass's four accepted rationalizations. Three held (fail-open while OSM opt-in; `globalThis` binding = 5th instance of a prod-proven pattern; constant-key + double try/catch). **One real MEDIUM surfaced and is now fixed:**

### Fixed: regeocode-miss could wipe a valid GEOCODED pin (`83798de`)
`resolveUpdateCoords` cleared a non-manual pin on **any** `regeocode: true` miss. Pre-#574 a forced refresh of an *unchanged* address actually hit OSM and got the same coords back; #574's throttle makes that miss **routine**, so under a saturated bucket a forced re-geocode silently destroyed a valid pin (data loss — reachable by API/3rd-party callers; the web UI never sends `regeocode`).

Fix: clear only genuinely-stale coords (address changed **and** re-derivation failed); preserve a MANUAL pin **or** a GEOCODED pin whose address is unchanged. Added a failing-first regression test (`location.test.ts`).

Gates after fix: **1210/1210** api tests · tsc 0 · boundaries OK · biome clean.

Issue #616, opened 2026-06-13
feat(operator): booking actions (substitute/cancel/status) + new-order red-dot — last canonical MVP gap


jackli921:
# Plan: Operator order management — substitute / cancel / status + new-order red-dot

## Context

This is the **last canonical-MVP gap**. The renter flow (search → book → confirm) is 100% done in the
Vite app, and operator steps 1–6 (login, locations, vehicles, per-vehicle pricing, insurance, add-on fees)
are all shipped. Only the operator's **接订单 (step 7)** and **管理订单 (step 8)** are partial — and almost
entirely on the **frontend**. The backend action endpoints are fully built and tested:

- `PATCH /bookings/:id/status` → `BookingService.updateStatus` (CAS-guarded; state machine
  `VALID_BOOKING_TRANSITIONS`: CONFIRMED→{ACTIVE,CANCELLED}, ACTIVE→{COMPLETED,CANCELLED}; returns 409 on race)
- `POST /bookings/:id/cancel` → **CONFIRMED-only**, returns `{ booking, cancellation }` (fee tier)
- `POST /bookings/:id/substitute` → gated `isOperatorRole`; `substitute()` enforces same-operator (404 else),
  AVAILABLE, **same pickup location**, **same ACRISS class via `sameAcrissClass()`**, re-snapshots price,
  atomic exclusion re-check, appends a `VEHICLE_SUBSTITUTED` `bookingEvents` row (audit = 留痕). The timeline
  already renders these events.

Scope is **web + one small read endpoint** (see §A) — no schema, no migration.

## Scope (in / out)

**In:** operator Actions panel (mark-active / mark-completed / substitute / cancel) on the trip-detail page,
a same-store same-ACRISS-class substitute picker, and a new-order red-dot badge on the operator Bookings nav.
**Out (side quests, stay focused):** manual booking (#589), operator vehicle utilization (#527), region
search (#394), renter docs (#515/#516), and email provider config (`RESEND_API_KEY` is ops config — the
`NotificationDispatcher`/`OPERATOR_BOOKING_ALERT` already exist; flag separately, don't build). Also out:
ACTIVE→CANCELLED (no-fee path) and a notification-ledger-based badge — both noted as follow-ups below.

### A. Substitute candidates — NEW endpoint (architect HIGH-1)
Do **not** filter `GET /vehicles` client-side: the response carries `classId` but **not** `acrissCode`
[...]

Issue #601, opened 2026-06-13
feat(api): KV cache + distinguishable 'pin pending' state for forward geocoding (#574 follow-up)


jackli921:
Follow-up to #574 (PR #591), which shipped the app-side rate-limit half. Two deferred pieces, both gated on the #304 Cloudflare account migration (KV/R2 not yet provisioned).

## 1. KV cache for geocode results
Cache `address → {lat,lng}` in Workers KV so repeated/identical lookups don't spend the 1-req/10s global budget (or a paid provider's daily quota). Keyed by normalized address; TTL'd. This is the second half of #574's original title.

## 2. Distinguishable 'pin pending' state (NOT a blind retry)
**Important scoping note from the PR #591 architect review (F6):** today a throttle-skip and a genuinely un-geocodable address are *both* persisted as `coordinateSource: null`. A deferred auto-retry has nothing to query against — it can't tell "throttled, retry will work" from "address is bad, retry won't help."

Before/with auto-retry, introduce a distinguishable pending state (e.g. a `coordinateSource` of `PENDING` or a nullable `geocodePendingAt`) so:
- bulk onboarding (#574 residual R2) can enumerate locations whose pin was *throttle-skipped* and re-geocode just those;
- the operator UI can surface "pin pending" vs "no pin found."

This needs a small schema migration (new enum value / column) — do it as part of this issue, not a blind retry-everything sweep.

## Context
- #574 / PR #591: `ThrottledGeocoder`, `GEOCODE_LIMITER` (1/10s), LocationIQ drop-in.
- Residual R2 (bulk silent pin-drop) and architect F6 (lossy failure signal — only a `console.warn` records the skip) are the drivers.

jackli921:
Piece 2/2 (distinguishable `PENDING` pin state) merged via #633 (squash `7bb1413`) into `marketplace-pivot`.

- Adds `coordinate_source = PENDING` (migration 0053) so a throttle-skip (#574, retryable) is distinguishable from a genuine geocode miss (`null`).
- Geocoder port returns a discriminated `GeocodeOutcome{ ok | notFound | throttled }`; service maps `throttled → PENDING`, `notFound → null`.
- Web: `LocationPinBadge` shows "pin pending" vs "no map pin" in operator-locations.

**Keeping this issue open** — piece 1/2 (KV geocode cache) is still outstanding and remains **blocked on #304**.

jackli921:
[...]

PR #632, opened 2026-06-13
feat: renter liability-disclaimer consent at checkout (#613)

jackli921:
Closes #613.

The canonical-MVP replacement for the dropped online document upload (#515/#516/#465, closed not-planned): a liability-disclaimer (免责声明) consent the renter acknowledges at checkout, recorded on the booking. Driver's-license / IDP verification moves to in-person at pickup.

## Backend (already on this branch, `e16218c`)
- Migration **0052** — `bookings.disclaimerAcknowledgedAt` (timestamptz, null) + `disclaimerTermsVersion` (text, null).
- Route gate: a `RENTER` booking POST without `disclaimerAccepted` → **400 `CONSENT_REQUIRED`** (staff / manual bookings exempt — only the renter self-serve path is gated).
- Service stamps consent server-side (`submitInTx`): server-derived timestamp + `DISCLAIMER_TERMS_VERSION='2026-06-13'`, null for non-renters.

## Web (this slice)
- `vite/bookings/api.ts` — `CreateBookingInput` gains required `disclaimerAccepted: boolean`; POST body sends it.
- `vite/reservation/PaymentStep.tsx` — native consent checkbox + terms text; **Reserve** gated on `!accepted`; submits `{...bookingInput, disclaimerAccepted: accepted}`. Prop typed `Omit<CreateBookingInput,'disclaimerAccepted'>` (`aria-describedby` links the terms to the checkbox).
- `vite/reservation/ReservationWizard.tsx` — `bookingInput` typed as the same `Omit`: consent is a payment-step gate, **not** a wizard selection, so the wizard can never assemble a booking input without the payment step supplying the flag.
- i18n `reservation.disclaimer.{label,terms}` in en / ja / zh.

**Why this shape:** consent state lives only in `PaymentStep` (colocated where it's used), and the `Omit` makes it impossible to construct a booking input without the payment step supplying the flag — single source of truth, illegal states unrepresentable. The client gate is UX; the server is the real authority (400 `CONSENT_REQUIRED`).

## Test plan
- [x] `PaymentStep.test.tsx` — Reserve disabled until the disclaimer is ticked; success POSTs `disclaimerAccepted: true`; 409 / 400 paths tick first then assert the choose-another message.
- [x] `ReservationWizard.test.tsx` — at the payment step Reserve is disabled, then enabled after ticking.
- [x] Gates: web suite 1057 passed · tsc 0 · i18n-parity green · biome clean.
- [ ] Manual browser smoke (not done).

PR #633, opened 2026-06-13
feat(api,web): distinguishable geocode PENDING pin state (#601 F6)

jackli921:
## Summary

Part of **#601** — delivers **piece 2 of 2** (architect finding **F6**: a distinguishable "pin pending" state). Piece 1 (Workers KV geocode cache) stays a follow-up, **blocked on #304** (KV not provisioned).

Today a geocode **throttle-skip** (#574 rate-limiter rejected the lookup — retryable later) and a **genuinely un-geocodable address** both persist `coordinate_source = null`. That's lossy: the bulk re-geocode can't tell which pins are worth retrying, and the operator UI can't distinguish "we'll fix this" from "this needs a manual pin". This adds a `PENDING` state to close that gap.

## What changed

- **shared**: `coordinate_source` enum gains `PENDING`. Migration `0052_add_coordinate_source_pending.sql` — additive, append-only `ALTER TYPE ... ADD VALUE` (structurally identical to the deployed `0047`).
- **api**: the `Geocoder` port now returns a discriminated union `GeocodeOutcome = {status:'ok',lat,lng} | {status:'notFound'} | {status:'throttled'}` (was `GeocodeResult | null`). `LocationService` maps `throttled → PENDING` (retryable) and `notFound → null` (genuine miss) via `failedTriple`, at all three sites (create, regeocode, address-changed). MANUAL pins and unchanged-GEOCODED pins are preserved — the #574 clobber regression stays locked down by test.
- **web** (web-only — `coordinateSource` was already on the wire from `GET /locations`): `OperatorLocation` DTO surfaces `coordinateSource`; new `LocationPinBadge` renders **only** for `PENDING` ("pin pending") and `null` ("no map pin"), nothing for GEOCODED/MANUAL. i18n added to en/ja/zh.

## Review

`code-reviewer` agent: **Ship** — no CRITICAL/HIGH. One MEDIUM fixed (vestigial `GeocodeResult` made load-bearing via the `ok` variant); two LOWs were no-change (badge already accessible; explicit `failedTriple` literal preferred). Migration confirmed safe (additive enum value, never read in-tx).

## Test plan

- [x] api unit: **1229 pass**, tsc clean
- [x] web: full suite **1049 pass**, tsc clean, i18n parity 848×3, biome clean
- [x] `db:verify` offline 3/3 (schema↔snapshot, journal↔disk, monotonicity)
- [ ] **journal↔DB sync + `test:integration`** — could not run locally (sandbox blocks the dev `.dev.vars` DB + docker); **validated by CI** `db-drift` + `test-and-build`. Reviewer confirmed `0052` matches the deployed `0047` pattern.

## Follow-ups

- #601 piece 1 (KV geocode cache) — keep #601 open or split a tracking issue; blocked on #304.
- Non-colliding with the #616/#619 order-actions swarm (touches only geocoding + operator-locations).


jackli921:
⚠️ **Migration `0052` collision — heads-up to avoid a double-merge break**

Both open PRs now branch from `mp@0051` and add a **different** `0052`:
- #632 (#613 consent): `0052_add_booking_disclaimer_consent.sql`
- #633 (#601 geocode): `0052_add_coordinate_source_pending.sql`

Both are CLEAN today only because neither has merged yet. **The first to merge takes `0052`; the second's `db-drift` job will then FAIL** on update-branch (duplicate sequence + journal `when` ordering).

**Whoever merges second must, before update-branch:**
1. Rename the `.sql` `0052_*` → `0053_*`
2. Bump its entry in `drizzle/meta/_journal.json` `when` to `max(previous) + 1` (out-of-order `when` is silently skipped by `drizzle-kit migrate` — see CLAUDE.md gotcha)
3. Regenerate the snapshot + `bun run db:verify` (3 green) before pushing

**Suggested order:** land the canonical P1 (#632 / #613) first; #633 (#601, a follow-up) renumbers to `0053`. Either order works — this is just to prevent both being squashed onto `mp` blind.

(FYI my #636 is harness-only, no migration — no collision.)

Issue #651, opened 2026-06-13
feat(marketplace): renter location search — wire NicoNico region picker into live search (build #394)


jackli921:
## Goal

Make the landing-page / search **location picker actually work**. Today the "Where" control is a static `<div>` reading `Osaka, Japan` (`packages/web/src/vite/landing/SearchWidget.tsx`) — no input, no handler; submit passes only `from`/`to` dates. This issue tracks **building** the renter location search end-to-end per the owner-approved design.

**Design (source of truth):** `docs/plans/2026-06-12-renter-location-search-niconico.md` (PROPOSAL v2, hardened 2026-06-13).
**Builds on:** #394 (region taxonomy *design* + Phases 1-3 foundation, currently unmerged on branch `feat/394-region-search`).

## Current state

- `#394` foundation (regions table, `locations.regionId`, `GET /regions`, recursive subtree filter on storefront + flat search) is **built but unmerged** — lives only on the foreign worktree `feat/394-region-search` (60 commits behind `marketplace-pivot`). Its migration is a stale `0048` that collides with mp's existing `0048`/`0053`.
- `#394` Phase 4 (the renter dropdown UI) was **never built** → this is the visible gap.
- Net-new beyond #394: region coords + suggestion engine, the operator→search loop guard, and the renter front door.

## Slices (vertical, strictly sequential — one PR each)

- [ ] **Slice 0 — Land #394 foundation.** Re-create `feat/394-region-search` Phases 1-3 on a fresh branch off `marketplace-pivot` (do NOT edit the foreign worktree). Regenerate the migration at the next free number (~`0054`); watch the journal `when`-order gotcha; `db:verify` green. While here, bound the availability scan (extend `AvailabilityFilters` with `locationIds`, push region-matched storefront ids into `findAvailableVehicles` in both search services). Effectively closes #394's build.
- [ ] **Slice 1 — Region coords + suggestion + backfill.** Migration adds `type/lat/lng/assignable/slug/status` to `regions`; seed region centers; pure `nearestAssignableRegion(regions, point)` in `@kuruma/shared`; idempotent backfill of existing locations; test: zero ACTIVE locations regionless.
- [ ] **Slice 2 — Operator→search loop.** Inject `RegionRepository` into `LocationService`; derive region from coords on create/update; **service-level guard** rejects effective-ACTIVE location with no region (422); validate `regionId` is assignable+ACTIVE; operator form shows suggested region + override.
[...]

PR #671, opened 2026-06-13
feat(api): land #394 region foundation + bound availability scan (#651 Slice 0)

### LINKED ISSUE #394: Renter location search — NicoNico-style hierarchical regions (hybrid v2, execution plan)
# Renter location search — NicoNico-style hierarchical regions (hybrid v2, execution plan)

**Status:** Owner-approved, execution-ready (revised 2026-06-13, hardened after adversarial review + a colleague scale pass).
**Epic:** #385 (marketplace MVP). **Builds on:** #391/#458 (storefront + flat search, merged), #531/#574/#601 (geocoding), and the unmerged region backbone on branch `feat/394-region-search`.
**Design doc (local, untracked — this issue is the canonical copy):** `docs/plans/2026-06-12-renter-location-search-niconico.md`

> **This replaces this issue's original "Approach A only" scope** (preserved below as an appendix). The v2 plan implements the hierarchical taxonomy (Approach A) as the *foundation* **and** folds in lat/lng (the old Approach B) as a *helper* for suggestion/distance/near-me — exactly the forward-compat hook the original note anticipated.

## TL;DR — is this the right call?

**Yes, validated.** An adversarial assessment compared the design against full-fledged rental sites (NicoNico, Times, Nippon, Orix + Rentalcars/Kayak/Airbnb/Turo) and against the real state of maps/geocoding in Japan:
- Renter search today is **date-only** (no location control at all), so this is the single most important missing axis, not a nice-to-have.
- The core bet — **region hierarchy as the source of truth; geocoding & maps as helpers** — is exactly what Japan's market leaders do, and the correct response to the fact that Japanese block-based addresses geocode poorly on *every* provider (incl. Google).
- A colleague review independently reached the same "right foundation" conclusion and added one scale requirement (region-aware availability query, §7) now folded in.

---

## 1. Goal

Give renters NicoNico-style hierarchical location search (**prefecture → city → area** dropdowns) that **works end to end**: when an operator registers a store, that store appears under the right area in renter search — automatically, without making operators classify geography from scratch. Layer tourist affordances (quick area chips, "📍 Near me", nearest-first ranking) on top of the same foundation. Scale nationwide by adding *data*, not schema.

## 2. The model

`marketplace-pivot` already implements everything *after* "pick an area" (storefront-first, epic #385). The missing axis is **WHERE**, and the right shape is a real region hierarchy, not a flat chip list:

| Layer | What | Source of truth |
|---|---|---|
| **Foundation** | DB-backed region tree (prefecture→city→area), recursive subtree filter | `regions` table + `locations.regionId` (#394) |
| **Shortcuts** | Quick chips (なんば 梅田 KIX 京都), "📍 Near me" | UI sugar → resolve to a `regionId` or a `lat/lng` point |
| **Ranking** | "~2.1 km · Namba" nearest-first *within* the chosen area's results | Haversine over store coords (#531) |

The hierarchy gives a true **filter**; distance is **ranking within** that filter. Chips are preset region selections. "Near me" skips the hierarchy and ranks by raw `lat/lng`.

## 3. What already exists vs. net-new

**Already built on `feat/394-region-search` (unmerged):**
- `regions` table — adjacency list (`parentId` self-ref, null=root), trilingual `nameEn/nameJa/nameZh`, `sortOrder`.
- `locations.regionId` — nullable FK + Zod validator + index; create/update routes thread it; FK→422 mapping.
- Recursive descendant resolution — `collectDescendantIds` (pure BFS, driver-agnostic; not a raw CTE on purpose).
- `GET /regions` — public, edge-cached 1h, flat list; client builds the cascade.
- `regionId` filter wired into **both** storefront-search and flat-search (filter to `regionId` subtree).
- Kansai seed (Osaka/Kyoto/Hyogo/Nara → city → area incl. KIX, Namba, Umeda) + integration tests.

**Net-new (the hybrid that closes the loop):**
- Extend `regions` with `type`, `lat`, `lng`, `assignable`, `status`, and a stable `slug`.
- A **region suggestion engine** (pure: nearest assignable region to a point).
- Operator location form: show the suggested region, allow override.
- Location service: derive a region from coords when none supplied; **reject an ACTIVE location with no region**.
- Backfill existing/seed locations from their `lat/lng`; assert no ACTIVE searchable location is unassigned.
- Renter: cascading **dropdown UI**, chips, "Near me", distance labels, map-centering.

### 4.1 `regions` (extend the existing table)
Current columns: `id`, `parentId`, `nameEn`, `nameJa`, `nameZh`, `sortOrder`, timestamps. **Add:**

| Column | Type | Why |
|---|---|---|
| `type` | enum `PREFECTURE \| CITY \| AREA` (extensible) | Label each dropdown level; decouples "level" from raw tree depth |
| `lat`, `lng` | double precision, nullable | Region center — powers map-centering, distance anchor, and the suggestion engine |
| `assignable` | boolean (default false) | Only assignable nodes (typically `AREA`) are valid store targets + suggestion candidates |
| `status` | enum `ACTIVE \| INACTIVE` (default ACTIVE) | Hide a region from dropdowns/suggestions without deleting it |
| `slug` | text, unique | **Stable** human ID for chips + URLs. Region `id` is `crypto.randomUUID()` regenerated on every seed → chips/URLs MUST use `slug` (`?region=namba`), never the UUID, or they break on re-seed. |

Region names stay **DB columns** (`nameEn/Ja/Zh`), not i18n message keys — so the renter UI picks by route locale and we never touch the i18n parity gate for region labels. **Trade-off:** adding a 4th locale later becomes a schema migration rather than a message-file edit. Acceptable at three locales.

### 4.2 `locations.regionId`
- **Nullable at the DB level** (migration safety: legacy rows, and the create path before suggestion runs).
- **Required by the service when effective `status = 'ACTIVE'`** (see §5). `regionId` is the **confirmed assignment and the only truth**; nearest-center is a *suggestion engine*, never a stored shadow field.
- **Must reference an `assignable`, ACTIVE region** — validated in the service (a Postgres `CHECK` cannot hold a subquery), → 422 not a raw FK 500.
- **`locations.status` is `ACTIVE | ARCHIVED` today — no `DRAFT`.** A newly created store is ACTIVE immediately and must resolve a region at create time. A draft-before-publish state is a separate future change, out of scope.

## 5. The operator → search loop (the part v1 missed)

Hooks into `LocationService.create/update`, which already resolves coordinates via the injected `Geocoder` (#531) — the suggestion runs right after.

1. **Operator registers/edits a store** → enters address (and/or drops a manual pin).
2. **Coords resolve** — `coordinateSource` ∈ `GEOCODED | MANUAL | PENDING` (#531/#601), or null when none captured. `PENDING` (throttled) means coords are unknown right now.
3. **Suggestion engine** (`nearestAssignableRegion(assignableRegionsWithCoords, point)`) proposes a `regionId`; the form shows it pre-filled. **No usable point → no suggestion:** when coords are null/`PENDING`, or the nearest assignable region is beyond a sanity radius (e.g. a Tokyo store while only Kansai is seeded), the engine returns nothing rather than a wrong guess, and the operator **must** pick a region manually before going ACTIVE.
4. **Operator confirms or overrides** via the prefecture→city→area dropdown.
5. **API create/update** accepts an explicit `regionId`, OR derives one from coords when absent.
6. **Guard — enforced in `LocationService`, NOT the route**, so operator + platform-admin + seed paths all obey it. If the location's **effective** status is `ACTIVE` (`data.status ?? existing.status` — a transition into ACTIVE can't sneak past it) and it still has no `regionId` after suggestion → **rejected (422)**. Fires on create AND update. `ARCHIVED` may stay unassigned.
7. **Backfill** (one-off idempotent script; Slice 1, **after** region centers seeded and **before** the Slice 2 guard deploys): assign every existing location from its `lat/lng`; stores beyond the sanity radius left null + reported (can't go ACTIVE until manually assigned). Test asserts no ACTIVE location is regionless.

**Suggestion vs. truth:** the function is stateless and advisory; we persist only the confirmed `regionId`. A boundary mis-snap is fixed by operator override. On an address edit the service re-suggests and the form surfaces the change ("region updated to Umeda"). Adding region **polygons** later changes only the suggestion engine — the search contract (filter by `regionId` subtree) is unchanged.

## 6. Renter search

- **Cascading dropdowns** (prefecture → city → area) built client-side from the cached `GET /regions` flat list.
- **Quick chips** (なんば 梅田 KIX 京都) = preset region selections referenced by **`slug`**, never the UUID.
- **Precedence (one anchor at a time):** the active "Where" is exactly one of — a selected region (dropdown/chip), a "Near me" point, or nothing. A selected region **always wins** over geolocation; "Near me" is opt-in, never default; on denial/unavailable, fall back to "nothing" (full list). Resolves the demo-from-outside-Japan footgun. "Near me" requires HTTPS (localhost OK in dev).
- **Results:** with a region chosen, filter storefronts to that subtree, then **rank nearest-first within the filtered set** with "~2.1 km" labels (Haversine from the anchor — region center or geolocation point — to each store). No region → today's behavior: **operator-name → name → id** stable sort. Null-coord stores rank last. "nearest-first" is a *within-page* ordering + label until the server-side `?sort=distance` follow-up (#439).
- **Store-grid card needs coords + an anchor:** today's `StorefrontCard` DTO has **no** `lat/lng` (only flat-search's `ResultLocation` does). So distance labels require (a) adding `latitude/longitude` to the `StorefrontCard` projection and (b) threading the anchor point into the grid. Optionally add `regionId` + a region breadcrumb to the card.
- **Map view (#458, built):** promote the toggle to co-equal. New: center the map on the chosen region's `lat/lng`; if null, fall back to the existing fit-all-pins viewport.

**Form caveat:** `StorefrontSearchForm` is deliberately **uncontrolled** (FormData on submit; dodges hydration flake #392). The Where control is a popover/dropdown resolving to a `regionId`/point merged into the submit nav — keep date inputs uncontrolled; the region selection is the one piece of local state.

## 7. Architecture & future-proofing

- **Nationwide = data, not schema.** New regions are rows; the adjacency tree + recursive filter already scale. MVP seeds **Kansai only**.
- **Tree stops at AREA.** Finer ("near my hotel/station") is free-text geocode (Phase 2), not thousands of landmark rows.
- **Distance sort is client-side, per page.** Fine at ~9 stores. Outgrows a page → server-side `?sort=distance` (#439), deterministic tiebreak for the cursor.
- **Region-aware availability query (REQUIRED before multi-operator scale).** Region filtering currently narrows only the *storefront list* — the **availability scan is not bounded**: with a region (not a single `pickupLocationId`) selected, `findAvailableVehicles(from,to,…)` passes no location filter and scans *all* available vehicles platform-wide, then discards out-of-region cars in memory. Fine at ~9 stores; a full-inventory scan per search at scale. **Fix:** extend `AvailabilityFilters` with `locationIds` and pass the region-matched storefront ids (already resolved after `findActiveStorefronts`) into `findAvailableVehicles` in **both** `StorefrontSearchService` and `FlatSearchService` (`location_id = ANY($ids)`, index on `vehicles.pickupLocationId`); in-memory path stays for tests. Cheap to land in Slice 0.
- **Recursion stays app-code BFS** (driver portability: neon-http `.rows` vs postgres-js array); revisit only at thousands of nodes.

## 8. Slices (vertical, TDD — strictly sequential)

Do not start Slice N+1 until Slice N's migrations are applied and `db:verify` is green.

- **Slice 0 — Land #394.** Bring `feat/394-region-search`'s commits onto a **fresh branch off current mp** (do NOT edit the foreign worktree). Its migration is a stale `0048` that collides with mp's existing `0048` (operator memberships; mp is past `0050`) — **do not rename the file**; rebase onto mp and regenerate (`bun run db:generate --name add_regions_and_location_region_id`) → delta at the next free number (≈`0054`), adding `regions` + the `regionId` FK to the existing `locations`. Mind the journal `when`-ordering gotcha. `db:verify` 3 green. **Also bound the availability scan** (§7): extend `AvailabilityFilters` with `locationIds` and pass region-matched ids into `findAvailableVehicles` in both services.
- **Slice 1 — Region coords + suggestion + backfill.** Migration adds `type/lat/lng/assignable/slug/status`. Then in order: (a) seed region centers + `type` + `assignable` + `slug`; (b) pure `nearestAssignableRegion(regions, point)` in `packages/shared/src/lib/` (assignable && ACTIVE && coords-present, Haversine, **sanity-radius cap**, tiebreak `sortOrder` then `id`, null point → null) shared by backfill/service/web; (c) one-off idempotent backfill; (d) test: zero ACTIVE regionless, far stores null + reported. Set `regionId` on seed locations.
- **Slice 2 — Close the operator loop.** Inject `RegionRepository` into `LocationService`; derive region from coords; **service guard** on effective status → reject ACTIVE-without-region (422) across operator/admin/seed; validate `regionId` assignable+ACTIVE (422); operator form shows suggested region + override dropdown + "region changed" hint.
- **Slice 3 — Renter front door.** Cascading dropdowns + chips (`slug`) + "Near me" + `lat/lng` on `StorefrontCard` + distance labels + map-centering; thread the chosen region (`slug`) through search→detail→back nav (#499 pattern).
- **Phase 2 (later):** free-text hotel/station search → public geocode endpoint behind #601 KV cache.

## 9. Test plan (mutation-resistant)

- **shared (unit):** `nearestAssignableRegion` — nearest assignable/ACTIVE/coords-present; ignores ineligible; null point → null; beyond radius → null; equidistant → deterministic tiebreak; null-coord-stores-last.
- **api (service):** derives region when absent; **effective-ACTIVE + no region → 422** (incl. status transition + admin/seed paths); ARCHIVED may be regionless; override beats suggestion; non-assignable/inactive → 422; PENDING/null coords → manual required. Backfill assigns in-radius; zero ACTIVE unassigned; far rows reported.
- **api (search, Drizzle):** `regionId` filter returns node + descendants; `StorefrontCard` projection **gains** `lat/lng` while still omitting operator internals (`licensePlate`, etc.). (Flat-search's `ResultLocation` is a separate DTO, not a regression target.)
- **web (component):** dropdown cascade from `/regions`; selecting an area pushes `slug` + re-filters/sorts; chip resolves to same `slug`; "Near me" denied → full list; chosen area overrides geolocation.
- **e2e (real-DB lane):** operator registers a Kansai store → renter picks that area → store appears nearest-first → opens → books. Plus: pick a far region → all stores still shown, ranked.

## 10. Decisions (owner-approved)

1. **Region assignment:** hybrid — auto-suggest from pin, operator overrides.
2. **Blank region:** not allowed for an effective-ACTIVE storefront (service-enforced; nullable at DB for migration safety). `ARCHIVED` may be unassigned. No `DRAFT` status.
3. **Backfill:** yes, via nearest-region with a sanity radius; tests prove every ACTIVE location has a region; far rows reported, not mis-assigned.
4. **Truth vs. suggestion:** `locations.regionId` is truth; nearest-center is only a suggestion engine.
5. **Breadth:** Kansai-only MVP; nationwide later by adding rows.
6. **Stable identifiers:** unique `slug`; chips/URLs use `slug`, not the UUID.
7. **Free-text hotel/station:** deferred to Phase 2 (behind #601).

## 11. Verified findings & risks (from the adversarial + geo assessment)

- **Geo verdict: SOUND, deliberately lightweight.** Region hierarchy as primary axis matches NicoNico/Nippon/Orix; renter never types an address; regionId (not coordinates) is truth. The geocoding layer (#531/#574/#601) is provider-neutral, throttled, cached, null-safe.
- **Map basemap DEFECT (must-fix, easy, separate from this design):** `PigeonMapAdapter` sets no `provider`, so pigeon-maps defaults to `tile.openstreetmap.org` and renders no attribution. Per the [OSMF tile usage policy](https://operations.osmfoundation.org/policies/tiles/), the public OSM tile server is best-effort with no SLA, requires attribution/caching/identification, and can be blocked — so **do not rely on it for a production/commercial service** (precise wording: it's not "forbidden," it's "not a production SLA source"). Fix: point at **GSI tiles** ([terms](https://maps.gsi.go.jp/development/ichiran.html) — real-time web/app use is OK with source attribution/link, "出典: 国土地理院"), a hosted tile provider, or self-hosted tiles + visible attribution. → filed as a fast-follow.
- **Operational:** production geocoding env must point at **LocationIQ or self-hosted Nominatim**, never the public `nominatim.openstreetmap.org`. Per the [Nominatim usage policy](https://operations.osmfoundation.org/policies/nominatim/) it's max 1 req/s, forbids autocomplete, requires caching, and explicitly points larger users to third-party/self-hosted. The code degrades gracefully (null geocoder if unset) but the prod value must be set deliberately.
- **Do NOT adopt Google Maps:** its [Geocoding policies](https://developers.google.com/maps/documentation/geocoding/policies) restrict caching/storage (except place IDs) and require map-displayed geocoding results to be shown on Google Maps — both conflict with our approach (we cache in #601 KV and display on pigeon-maps); also a cost trap at this scale. Future geocode swap, if ever needed: `jageocoder` (offline, Japan-specific), not another OSM-derived API.

## 12. Out-of-scope fast-follows (file separately — required for a "full-fledged" site, but not this feature)

- [ ] #657 — **Cancellation policy shown before booking** (TABLE-STAKES) — the tiered 72h/48h/24h/same-day policy currently appears only *after* booking; add it to `ConfirmStep` pre-pay.
- [ ] #658 — **Renter class/category filter UI** (TABLE-STAKES) — the API supports ACRISS filtering; the search form is date-only and never lets a renter set a class.
- [ ] #659 — **Multi-photo gallery** on card/detail (TABLE-STAKES) — both render only `photos[0]`; the data is an array.
- [ ] #660 — **Map basemap tile source + attribution fix** (see §11) — one-file change in `PigeonMapAdapter`.

Safe to defer (tracked, not blocking): free-text airport/station search (Phase 2), server-side distance sort (#439), price/seat/feature filters, ratings & reviews, one-way drop-off.

---

## Appendix: original design note (2026-05-29)

> Preserved verbatim for history. The v2 plan above supersedes the "Approach A only" MVP decision by implementing A as the foundation and folding in B (lat/lng) as a helper.

- **Two approaches considered:** (A) hierarchical region taxonomy (NicoNico-style) — chosen for MVP; (B) lat/lng + distance radius — deferred.
- **MVP decision:** Approach A with a forward-compat hook for B: `regions` table with `parent_id`; `locations.region_id` FK; search filters by node + recursive descendants; pre-seed `locations.latitude/longitude` now (unused) to avoid future migration churn.
- **Out of scope (post-MVP):** "near me", map view, free-text region search, operator self-serve region creation. *(v2 now designs the first three.)*
- **Original acceptance criteria:** cascading dropdowns (prefecture→city→area); selecting a level filters storefront cards to node + descendants; operator location form has a region picker; `regions` seeded with ≥ Osaka's areas; `locations` has `region_id`, `latitude`, `longitude`.
- **Engineering concepts:** Hierarchical Taxonomy (tree via parent_id + recursive descendants); Forward-Compatible Schema (lat/lng added early).

jackli921:
Slice 0 of #651: land the #394 region taxonomy foundation on `marketplace-pivot`, then bound the availability scan. Two commits.

## 1. Land the #394 region foundation (`3accdcd`)
Re-applies #394 Phases 1-3 (built unmerged on `feat/394-region-search`) onto mp via a **3-way merge, not a cherry-pick** — mp had diverged (geocode-on-save `LocationService`, `AppOverrides`, `StorefrontDetailService`), so the search services / `location.ts` / `index.ts` were hand-merged to keep BOTH mp's #531 lat/lng+coordinateSource AND the foundation's `regionId`.
- `regions` table (adjacency list, trilingual names) + `locations.regionId` FK
- `GET /regions` (public) + recursive subtree filter on storefront-search and flat-search
- `regionId` settable on location create/update; bad id → 422 (not FK 500)
- Kansai seed + region/location integration tests
- Migration regenerated as **0054** (foundation's stale 0048 collided with mp's 0048/0053)
- Code-review **HIGH** folded in: `collectDescendantIds` BFS gains a visited-set — the `regions.parentId` self-FK has no DB cycle constraint, so a cyclic row would infinite-loop the public, unauthenticated search path (CPU-limit DoS). Adds cycle/dedup unit tests.

## 2. Bound the availability scan to in-region storefronts (`ea43ddc`, §1c)
A region selection previously narrowed only the storefront list; `findAvailableVehicles` still scanned **every vehicle platform-wide** and discarded out-of-region cars in memory.
- New `AvailabilityFilters.locationIds` (alongside the singular `locationId`).
- Both `StorefrontSearchService` and `FlatSearchService` pass the region-filtered storefronts' ids into the scan.
- Drizzle uses `inArray` over `idx_vehicles_pickupLocationId` (empty set guarded → match nothing); in-memory path mirrors.
- `StorefrontDetailService` is single-storefront and unaffected.

## Test plan
TDD (RED→GREEN per behavior):
- in-memory + Drizzle integration `locationIds` filtering, incl. the empty-set "no in-region storefront" case
- a port-contract test in each search service proving a region selection bounds the scan (asserts the in-region storefront ids reach the repo) rather than scanning the whole fleet

Gates (local, disposable PG): shared 474 · api unit 1290 · api integration 228 · typecheck · biome · boundaries OK · 47 FKs indexed · db:verify 4/4.

## Heads-up — pre-existing flake (not from this PR)
The api unit lane has a test-isolation flake in 3 route files (`auth-middleware`, `auth-session`, `rate-limit`): they fail intermittently **only** under the full parallel run, and pass 25/25 in isolation, on the clean base, and on this branch. If the unit lane goes red there, a re-run clears it — it is unrelated to region/availability code. Filing a follow-up to fix the isolation.

Closes #394
Refs #651


jackli921:
Follow-up for the pre-existing unit-lane flake noted above: #672.

jackli921:
Heads-up from the #676 train (conceding the region dup to this PR as canonical 👍): the body says `Closes #394`, but this is Slice 0 only — #394 Slices 1–3 (region coords/suggestion, operator loop, renter front door) still follow. Consider downgrading to `Refs #394` so #394 stays open after merge. FYI the web region picker (StorefrontSearchForm + i18n) from the now-closed #675 is preserved on branch `feat/394-region-search-slice0` (`e6f58abb`) and can be re-landed as a separate #651 PR once your foundation is on trunk.

PR #675, opened 2026-06-13
feat(#394 Slice 0): region taxonomy backbone + region-bounded availability scan

jackli921:
## Slice 0 of #394 (renter region search)

Lands the **region taxonomy backbone** onto `marketplace-pivot` and closes the one scale gap the foundation left open. This is **Slice 0 only** — Slices 1–3 (region coords/suggestion/backfill, operator loop, renter front door) follow. Per issue #394 §8.

### What's in this slice
**Foundation (cherry-picked from `feat/394-region-search`, conflicts resolved onto current mp):**
- `regions` table — adjacency list (`parentId` self-ref), trilingual `nameEn/Ja/Zh`, `sortOrder`. Lives in `db/regions.ts` (keeps `schema.ts` under the 800-line cap).
- `locations.regionId` — nullable FK + Zod validator + index; create/update routes thread it (FK violation → 422).
- `GET /regions` — public, edge-cached, flat list; the client builds the cascade.
- `regionId` filter on **both** storefront-search and flat-search (node + recursive descendants). Descendant resolution is a shared **app-side BFS** over a single `findAll()` (`region-tree.ts`) — deliberately not a SQL CTE (driver result-shape portability).
- Kansai seed (Osaka/Kyoto/Hyogo/Nara → city → area incl. KIX, Namba, Umeda).

**Migration:** regenerated as `0054_add_regions_and_location_region_id` against current mp (the foreign branch's stale `0048` was discarded). Purely additive; journal `when` is post-`0053`.

**Net-new — region-bounded availability scan (§7):** region filtering previously narrowed only the storefront *list*; `findAvailableVehicles` still scanned the **whole fleet** and dropped out-of-region cars in memory — a full-inventory scan per search at scale. Now `AvailabilityFilters.locationIds` carries the region-matched storefront ids into the scan in **both** services.
- **Invariant — never widen:** an EMPTY set short-circuits to zero vehicles. Drizzle returns early *before issuing any query*; in-memory excludes all. The old single `locationId`/`pickupLocationId` filter is subsumed (findActiveStorefronts already narrows to it).
- **Behavior-preserving:** results are identical — only the scan is tighter.

### Test plan
- **Unit (TDD RED→GREEN):** in-memory repo `locationIds` + empty cases; `vi.spyOn` in both service suites proves the scan receives `locationIds` (the load-bearing never-widen assertion — the bug is invisible at the result level).
- **Integration (real pg):** Drizzle `locationIds` filter; empty-set test asserts **no query is issued** (mutation-resistant guard test).
- **Gates green:** tsc, biome, lint:size, lint:boundaries, db:verify 4/4 (fresh DB, 55 migrations), api unit 1292, integration 28 files/228, web 1103, shared 474.

### Review
Multi-agent adversarial review → 3 findings, all LOW, all addressed: two stale "WITH RECURSIVE CTE" comments corrected to describe the real app-side BFS; the Drizzle empty-scan comment corrected (`inArray([])` compiles to `false`/zero-rows in drizzle 0.45.2 — the early return is a query-skip + driver-robustness guard) and its integration test hardened to pin the guard.

Part of #394. Does **not** close it — Slices 1–3 remain.


jackli921:
Closing as a duplicate of #671 — both independently implement #394 Slice 0 (region taxonomy foundation re-land + region-bounded availability scan), and they're mutually exclusive (both create the `regions` table and migration `0054`).

Keeping **#671** as the canonical Slice 0 because:
- It was opened first (14:26 vs 14:36) and is **mergeable**; #675 is currently CONFLICTING against `marketplace-pivot`.
- #671 carries the foundation code-review **HIGH** fix that #675 lacks: a visited-set **cycle guard** in `collectDescendantIds` (`region-tree.ts`) + its test. Without it, a single cyclic `regions.parentId` row infinite-loops the **public, unauthenticated** search endpoint (CPU-exhaustion DoS). #675's `region-tree.ts` has no such guard.
- Both have 5/5 CI green; the difference is mergeability + the DoS fix.

No reflection on the work here — it's a near-equivalent build; this is just swarm de-duplication. Leaving this branch/worktree for its owner to tear down.

jackli921:
👋 Heads-up to whoever's driving this branch (`feat/394-region-search-slice0` / worktree `kuruma-394-slice0`): **your Slice 0 work here is being closed** as a duplicate of #671, which is the canonical Slice 0 going forward.

Nothing wrong with the build — it's a near-equivalent implementation; this is just swarm de-duplication, and #671 edged it out on three counts: opened first, currently mergeable (this branch is CONFLICTING vs mp), and it carries the cycle-guard DoS fix in `region-tree.ts` that this branch's version lacks.

No action needed from you on the code. When you get a chance, you can tear down the `kuruma-394-slice0` worktree + `feat/394-region-search-slice0` branch (I left them untouched). Thanks for the parallel effort. 🙏

Issue #701, opened 2026-06-13
epic(tech-debt): codebase health hardening — post-MVP audit 2026-06-13


jackli921:
Tracker for the 39 atomic issues from the deep multi-agent codebase health audit of `marketplace-pivot`.

Full report (problem + approach + acceptance per issue, dependency waves): `docs/2026-06-13-codebase-health-audit.md`

> Functionally sound and feature-complete, but launch-hardening is gated by one snapshot-invisible double-booking guard, a wrong-stack .env.example, and a half-finished Next→Vite migration leaving ~12k LOC of dead UI invisible to tooling.

**Shape:** 2 P0 · 7 P1 · 13 P2 · 17 P3 · (21 S / 15 M / 3 L) · 31/39 parallel-safe · 3 dependency waves.

## Themes
- Half-finished Next->Vite migration: three competing web feature trees, ~12k LOC of dead Next UI still type-checked and shipped in deps/lockfile, with knip blind to the live SPA so rot accumulates undetected
- Latent authz/data-loss hardening gaps that pass all current checks: snapshot-invisible double-booking exclusion constraint, message-translate route 500-not-401, tenant-unblind maintenance-logs read, wrong-stack .env.example
- Architecture-boundary erosion governed by docs not lint: routes calling repos directly, concrete repos built outside the composition root, undocumented carve-outs (tx factories, web Auth.js DB access) that future contributors cannot distinguish from drift
- API service/middleware god-files and enum/union hand-mirroring across files: SRP debt plus drift footguns that compile green but break sends or correctness
- Type-safety seams that trust the wire: unvalidated API-response casts, non-discriminated event union force-casts, scattered ?? vs || and non-null-assertion hazards
- Performance correctness on hot paths bounded by current scale: N+1 in substitution candidates, unbounded admin-revenue scan, O(total)-per-page search pagination
- Repo hygiene debt: 147 undeleted squash-merged branches, 20 ephemeral handoff docs in the tracked tree, dead scripts and knip false-positives blocking a hard dead-code gate

### Wave 1 (27) — no deps, fully parallel
- [x] #702 `P0/S` ∥ — fix(db): add db-verify check asserting critical custom DB objects exist (double-booking exclusion, triggers, idempotency index)
- [x] #703 `P0/S` ∥ — docs(env): rewrite .env.example to the real Auth.js v5 stack and delete pre-pivot Supabase keys
[...]

PR #800, opened 2026-06-14
fix(api): explicit bare /availability auth mount (#739)

### LINKED ISSUE #739: fix(api): add explicit bare /availability mount to the auth allow-list
**Priority:** P3 · **Effort:** S · **Dimension:** Security & authorization · **Parallel-safe:** yes

## Problem
index.ts mounts app.use('/availability/*', requireAuth()) but the bare GET /availability has no explicit gate; it currently 401s because Hono's * wildcard also matches the bare path (deterministic, confirmed empirically). By contrast /customers and /documents were given BOTH the bare and /* mounts explicitly. No real authz gap (401 confirmed; availability is also public via /storefronts and /search and non-staff get conflict details stripped) — purely a consistency nit to remove reliance on wildcard matching.

## Proposed approach
Add an explicit app.use('/availability', requireAuth()) line next to the existing /availability/* mount, matching the bare+wildcard pattern already used for /customers and /documents.

## Acceptance criteria
- [ ] index.ts has both /availability and /availability/* requireAuth mounts
- [ ] Unauthenticated GET /availability returns 401
- [ ] Authenticated GET /availability still works
- [ ] No other route behavior changes

## Affected files
- `packages/api/src/index.ts`

---
Part of #701 · Source: codebase health audit 2026-06-13 (`docs/2026-06-13-codebase-health-audit.md`)

jackli921:
Part of the #701 audit epic.

## Problem
`index.ts` mounts `app.use('/availability/*', requireAuth())` but the bare `GET /availability` had no explicit gate — it 401s only because Hono's `*` wildcard incidentally matches the bare path. `/customers` and `/documents` were given both the bare and `/*` mounts explicitly; `/availability` was not. No real authz gap (401 confirmed), purely a consistency/robustness nit.

## Change
Add an explicit `app.use('/availability', requireAuth())` next to the wildcard mount, mirroring the established bare+wildcard pattern. Guard tests in `auth-middleware.test.ts` (real composed `createApp()`) assert unauthenticated bare `/availability` → 401 and authenticated → 200.

## Test plan
- [x] `@kuruma/api` — 1439/1439 (incl. 2 new guard tests)
- [x] tsc · biome lint · import-boundaries — clean

Closes #739

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -299,6 +299,9 @@ export function createApp(overrides?: AppOverrides) {
   app.use('/vehicles/*', requireAuth())
   app.use('/bookings/*', requireAuth())
   app.use('/availability/*', requireAuth())
+  // Bare path explicit (don't rely on the wildcard matching it) — same
+  // bare+wildcard pattern as /customers and /documents below (#739).
+  app.use('/availability', requireAuth())
   app.use('/threads/*', requireAuth())
   app.use('/messages/*', requireAuth())
   app.use('/customers/*', requireAuth())
--- a/packages/api/tests/routes/auth-middleware.test.ts
+++ b/packages/api/tests/routes/auth-middleware.test.ts
@@ -150,4 +150,26 @@ describe('auth middleware', () => {
     expect(res.status).toBe(401)
     process.env.PARTNER_API_KEY = undefined
   })
+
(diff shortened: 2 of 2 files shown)


jackli921:
Duplicate of #802, which already landed the explicit bare /availability mount (#739) on marketplace-pivot with its own test coverage. Closing; #739 is resolved by #802.

Issue #819, opened 2026-06-14
refactor(api): extract VehicleService — move stranded business policy out of routes/vehicles.ts


jackli921:
Deferred half of #692 (the "extract later" of the agreed split-now/extract-later plan).

#692 reconciled the routes→repositories lint with reality: `regions`/`stats` are genuinely-thin reads and are now a permanent, documented sanction (they may DI only their `*Repository` interface). **`vehicles.ts` is the exception** — it is NOT thin. It's a 257-line CRUD route with real business policy stranded in the handler:

- `PATCH /vehicles/bulk-status` — existence pre-check + RETIRED-status guard
- `PATCH /vehicles/:id` — patch/existing merge, "at least one rate" rule, min≤max rental-hours rule
- `POST`/`PATCH` — FK-violation → 422 message mapping (`fkViolationMessage`), unique → 409
- operator-id resolution for writes

That logic can't be reused (e.g. a Trip.com path) or unit-tested without spinning up HTTP.

## Fix
Extract a `VehicleService` following the #712 pattern (MessageService / UserDirectoryService / AvailabilityService): discriminated-union result, `parseId`/`parseBody` stay at the HTTP boundary, the route becomes parse → call service → map result. Keep behavior identical (existing vehicle route tests are the net; add service unit tests).

Then close the carve-out:
- Remove `routes/vehicles.ts` from `EXTRACTION_PENDING_ROUTES` in `packages/api/scripts/check-import-boundaries.ts` (it joins the enforced set).
- Update the AGENTS.md "API Layer Architecture" note (only `regions`/`stats` remain sanctioned).
- Drop the now-obsolete extraction-pending boundary tests.

## Refs
- #692 (split-now/extract-later; lint + docs + sanction landed there)
- #712 (the route→service extraction pattern to follow)
- #726 (the routes→repositories boundary lint)

jackli921:
Done — VehicleService extracted (route→service layering gap closed for vehicles). PR #848 squash-merged `38bcfc39` to marketplace-pivot. code-reviewer: SHIP (one LOW addressed). Gates green: tsc×3, api 1500 unit, web 787, shared, lint:boundaries (vehicles now fully enforced — EXTRACTION_PENDING carve-out removed), e2e-real-db.

Issue #726, opened 2026-06-13
chore(api): forbid repository imports inside routes/ (move request filter types to shared/service layer)


jackli921:
**Priority:** P3 · **Effort:** M · **Dimension:** Architecture & module boundaries · **Parallel-safe:** NO (sequential)

## Problem
11 route files import repository interface types from ../repositories/types (VehicleFilters, BookingFilters, AvailabilityRepository, etc.), so data-layer filter/entity shapes leak into the controller layer and a repo-interface change ripples into HTTP handlers. The boundary linter deliberately whitelists type-only repo imports in routes for DI, so this is codified-but-inconsistent with AGENTS.md prose ('routes... Never repositories'). Type-only, zero runtime effect — a boundary-clarity nit. (Note: the DI-injected routes with no service layer legitimately need the repo interface as their constructor contract; resolving that overlaps with the MessageService/UserDirectoryService/AvailabilityService extraction.)

## Proposed approach
Move request-shaped filter types (VehicleFilters, BookingFilters, etc.) used by service-backed routes into the service layer or @kuruma/shared, and import them from there. For routes that genuinely DI a repo with no service (handled by the service-extraction issue), leave the constructor-contract import until that lands. Then tighten check-import-boundaries.ts / AGENTS.md so the prose and the codified rule agree, forbidding repository imports in routes except the documented DI carve-out.

## Acceptance criteria
- [ ] Service-backed routes import filter types from shared/service, not ../repositories/types
- [ ] check-import-boundaries.ts and AGENTS.md no longer contradict each other on the routes->repositories rule
- [ ] lint:boundaries passes
- [ ] api typecheck passes

## Affected files
- `packages/shared/src/types/`
- `packages/api/src/routes/bookings.ts`
- `packages/api/src/routes/add-ons.ts`
- `packages/api/src/routes/insurance-options.ts`
- `packages/api/src/routes/fee-schedules.ts`
- `packages/api/scripts/check-import-boundaries.ts`

---
Part of #701 · Source: codebase health audit 2026-06-13 (`docs/2026-06-13-codebase-health-audit.md`)
**Depends on:** "refactor(api): extract MessageService/UserDirectoryService/AvailabilityService so routes stop embedding business logic"

jackli921:
Blocked by: #712

jackli921:
[...]

Issue #712, opened 2026-06-13
refactor(api): extract MessageService/UserDirectoryService/AvailabilityService so routes stop embedding business logic


jackli921:
**Priority:** P2 · **Effort:** L · **Dimension:** Architecture & module boundaries · **Parallel-safe:** yes

## Problem
messages.ts, users.ts, and availability.ts have no service layer and embed genuine business rules in the HTTP controller: messages.ts has an idempotentCreate (UNIQUE_VIOLATION race) helper + participant authz; users.ts has the allowedUserIds name-harvest-prevention authz; availability.ts has the staff-vs-renter conflict-redaction rule. These rules can't be unit-tested without spinning up Hono and can't be reused by another caller (Trip.com path / future job), and it defeats the documented routes->services->repositories direction. (vehicles.ts/stats.ts are thin passthrough reads and are out of scope for this issue.)

## Proposed approach
Introduce MessageService (owns idempotent create + participant authz), UserDirectoryService (owns allowedUserIds filtering), and AvailabilityService (owns staff-vs-renter conflict redaction). Wire them in index.ts (composition root) and have each route inject the service and call a single method, leaving routes at parse->call->respond. Keep the existing repo interfaces. This is L — land as three focused PRs (one service each) if it would touch more than ~8 files together; each service+route+wiring is independently mergeable.

## Acceptance criteria
- [ ] messages.ts, users.ts, availability.ts contain no idempotency/authz/redaction logic — only parse, service call, respond
- [ ] Each new service is unit-tested without Hono (idempotency race, allowedUserIds filtering, conflict redaction)
- [ ] Behavior is unchanged at the HTTP boundary (existing route behavior preserved by tests)
- [ ] lint:boundaries still passes

## Affected files
- `packages/api/src/routes/messages.ts`
- `packages/api/src/routes/users.ts`
- `packages/api/src/routes/availability.ts`
- `packages/api/src/services/`
- `packages/api/src/index.ts`

---
Part of #701 · Source: codebase health audit 2026-06-13 (`docs/2026-06-13-codebase-health-audit.md`)

jackli921:
Slice C (AvailabilityService) merged via #768 (62a39ad1). Slices A (MessageService) + B (UserDirectoryService) still to follow — issue stays open.

jackli921:
All three slices merged to marketplace-pivot — routes no longer embed business logic:
- Slice C — AvailabilityService: #768
- Slice B — UserDirectoryService: #773
[...]

Issue #692, opened 2026-06-13
refactor(api): close route→service layering gap (routes injecting repos directly)


jackli921:
AGENTS.md says 'Routes import services and routes/helpers.ts only. Never repositories.' But `scripts/check-import-boundaries.ts:75` **explicitly allows** `from '../repositories/types'` (interfaces) and only blocks concrete drizzle/in-memory imports — so the enforced rule is weaker than the documented one.

**5 routes take repositories directly and strand business logic in the handler:**
- `routes/messages.ts:34` — idempotent-create logic at `:12-31`
- `routes/users.ts:37` — renter name-harvest authz policy at `:18-30`
- `routes/availability.ts:6` — staff-vs-renter conflict projection at `:29-35`
- `routes/vehicles.ts:36-39`, `routes/stats.ts:13` — thin reads

That logic can't be reused (e.g. a Trip.com path) or unit-tested without HTTP.

**Fix:** extract `MessageService` + `UserDirectoryService` for the two with real policy; for the genuinely-thin reads (stats/availability), sanction the pattern explicitly and tighten the lint so prose and enforcement agree. Effort: M.

Pattern — SRP. Heuristic: if a route reads a repo result then branches on a business rule, the branch belongs in a service.

---
_Source: 2026-06-13 maintainability audit (`docs/2026-06-13-maintainability-audit.md`), 5-agent review of `marketplace-pivot`@87f5b327. All findings grounded in verified `file:line`._

jackli921:
**Atomic unit** · Scope: `routes/{messages,users,availability}.ts` + new `services/*` + `scripts/check-import-boundaries.ts` (+ service wiring in `index.ts`) · Depends on: none · **Coordinate with #691** (shared route files) and **#693** (both touch `index.ts`). Can be split further into independent `MessageService` vs `UserDirectoryService` PRs (disjoint route files) if you want two agents.

jackli921:
Starting. Reconciled against trunk `b5bb89d6`: 3 of the 5 routes this issue named (messages/users/availability) were already extracted to services in #712, and the lint↔prose contradiction was already tightened by #726 — leaving a **transitional carve-out for regions/stats/vehicles** (`check-import-boundaries.ts:142`).

[...]

PR #821, opened 2026-06-14
refactor(api): sanction thin-read DI-repo routes; scope vehicles for extraction (#692)

jackli921:
## What
Closes the routes→repositories "prose vs enforcement" gap for the genuinely-thin routes, and precisely scopes the rest.

On trunk, 3 of the 5 routes #692 named (messages/users/availability) were already extracted in #712, and #726 tightened the lint — leaving a **transitional carve-out for regions/stats/vehicles**. This PR resolves the carve-out per the maintainer's split decision:

- **Sanction `regions`/`stats`** as a *permanent* thin-read exception — but only for their `*Repository` interface. An entity/filter import now fails the lint, so a thin route can't grow fat behind the carve-out.
- **`vehicles` is NOT thin** (257-line CRUD with real stranded policy) → kept as the lone *extraction-pending* route; filed **#819** to extract a `VehicleService` (4th sibling to #712 A/B/C).
- Reconciled **AGENTS.md** prose with the enforced rule.

## Tests (TDD)
18/18 boundary tests green (RED→GREEN on the two new `*Repository`-only invariants); `lint:boundaries` OK on the full api tree; api typecheck clean. Behavior-preserving.

## Scope
Part of #692 (stays open until #819 lands). Follow-up: #819.

> Status: needs code-review (drill Phase 4/5) before merge.


jackli921:
## Architect review — SHIP-WITH-FIXES (2 small doc fixes)

Independent architect pass on the thin-read sanction. **Verdict: mergeable** — behavior-preserving (no route source changed), and the `*Repository`-suffix heuristic is currently airtight: every `*Repository` symbol in `repositories/types.ts` is an `interface` contract, so there's no entity-laundering vector. Multi-line imports are correctly handled (the collapse step at `check-import-boundaries.ts:65-82` feeds the per-line scan, so a wrapped entity import into regions/stats is still flagged — verified empirically).

Credit where due: `75259f5e` ("hoist carve-out routes to a named registry; pin alias-direction") already landed the **policy-table/registry** refactor and alias hardening — that was going to be my main LOW recommendation (replace the `isSanctionedThinReadRoute`/`isExtractionPendingRoute` boolean pair before a 3rd category arrives). Nice — done.

### Two remaining MEDIUM findings — both in `AGENTS.md` (untouched by `75259f5e`):

**1. "no business policy" over-reaches → it becomes precedent for inlining the next check.**
`routes/stats.ts:17-20` runs an inline `X-API-Key` `timingSafeEqual` gate before the read. So "regions/stats are read-only with no business policy" isn't quite true for stats. The carve-out is still correct (a transport-level auth check is a legitimate *route* concern, not domain logic) — but the prose should say so, or the next reader cites "stats does auth inline and that was sanctioned" to justify a fatter route.

**2. No documented graduation trigger for sanctioned routes.**
`vehicles` has #819 as its "extract a service" tripwire; regions/stats have no symmetric rule for *when a sanctioned route loses its sanction*. One line closes the OCP gap (today, adding policy to regions means editing the carve-out; the rule makes it extract a service instead).

Suggested `AGENTS.md` row replacement (drop-in):

> ...Two scoped carve-outs (#692): **sanctioned thin-read routes** (`regions`, `stats`) carry no *domain* policy — a pass-through service would be anemic — so they may DI *only* their `*Repository` interface from `types.ts` (an entity/filter import is query-shaping that belongs in a service, and stays blocked; transport-level auth such as `stats`' API-key check is a route concern and stays put). A sanctioned route that later grows domain logic (authz scoping, response shaping, orchestration) must **graduate to a service** — the `vehicles`/#819 path — not widen its carve-out. The lone **extraction-pending route** (`vehicles`) keeps a path-wide `types.ts` exemption transitionally until its `VehicleService` lands (#819).

(Mirror the same "no DOMAIN policy / transport-auth stays put / graduate don't widen" clarification in the `check-import-boundaries.ts` sanctioned-route comment so prose ↔ enforcement stay reconciled.)

_Posted as a comment rather than a commit: this worktree is being actively iterated by a parallel session (HEAD moved `902f1588`→`75259f5e` mid-review), so I'm not editing it directly._

Issue #643, opened 2026-06-13
Operator-only gate on PATCH /bookings/:id/status (renter can self-advance via raw API)


jackli921:
**Found during #616 code review (pre-existing on `marketplace-pivot`, not a regression from PR #642).**

\`PATCH /bookings/:id/status\` (`packages/api/src/routes/bookings.ts`) and \`POST /bookings/:id/cancel\` apply only row-scoping (\`bookingReadScope\`), no role gate. A RENTER scoped to their own booking can therefore drive \`CONFIRMED → ACTIVE → COMPLETED\` via the raw API — pickup/return are physical operator events, so these transitions are plausibly operator-only.

By contrast `/substitute` and `/substitution-candidates` (added in #642) ARE role-gated (`isOperatorRole` → 403).

**Decision needed:** are renter self-service status transitions intended?
- If no: add `if (!MANAGEMENT_READ_ROLES.has(ctx.role)) return fail(...)` to `/status` (mirror `/substitute`).
- Renter self-*cancel* is plausibly intended-open by design — confirm separately.

```
Learn: Broken Access Control (missing function-level authz)
A state-changing endpoint that relies only on row-scoping (not a role gate) lets any scoped caller perform a privileged-role action.
At scale a renter scripts their own COMPLETED transition, skewing operator dashboards and any settlement keyed off status.
Heuristic: every state-changing route needs BOTH "who can do this verb" AND "which rows" — row-scoping alone is not authorization.
```

jackli921:
Fixed in PR #646 (squash → marketplace-pivot). Closing manually — base is marketplace-pivot, not the default branch.

PR #850, opened 2026-06-14
test(api): lock CANCELLED fee asymmetry + clarify the intentional split (#679)

### LINKED ISSUE #679: Dedup dual path to CANCELLED (/status updateStatus vs /cancel) + clarify fee semantics
Split out of #664 (PR #674). Two routes reach CANCELLED: PATCH /bookings/:id/status (updateStatus) and POST /bookings/:id/cancel. #664 kept CANCELLED in STATUS_TRIGGER to fire renter email on operator-cancel-of-ACTIVE; operator-cancel charges no renter fee. Route dedup + fee semantics is a product decision — track here. Not blocking.

jackli921:
Closes #679

## The reframe
The issue title says "dedup dual path to CANCELLED." On investigation the two paths are **intentionally divergent, not accidentally duplicated** — a naive "unify both paths" refactor would be a bug, not a cleanup. So this PR *clarifies and guards* the split instead of merging it.

| | renter `POST /cancel` | operator `PATCH /status` -> CANCELLED |
|---|---|---|
| Fee | **always** tiered (`calculateCancellationFee`) | **never** — `cancellationFee` NULL |
| Authz | renter self-cancel (row-scoped) | `MANAGEMENT_READ_ROLES` gate (#643) |
| Event | `BOOKING_CANCELLED` (+fee) | `STATUS_CHANGED` |
| Valid from | CONFIRMED only | CONFIRMED / ACTIVE |
| Returns | `{ booking, cancellation }` | `{ booking }` |

## Fee semantics (the product question in the issue) — decided
Operator cancellation stays **free**: operator non-delivery is the operator's fault, so the renter is made whole (full refund), never penalised. Charging a fee on the operator path would mean the renter loses money because the *operator* failed — the opposite of standard marketplace behavior. Confirmed with the owner this session.

## What changed (no behavior change)
- **Guard test** (`booking.test.ts`): operator `updateStatus(-> CANCELLED)` leaves `cancellationFee` NULL, returns no fee breakdown, and appends `STATUS_CHANGED` (not the fee-bearing `BOOKING_CANCELLED`). Verified to go RED if a "unify" refactor attaches a fee to the operator path.
- **Docs**: fault-based fee rule on `cancel()` and `updateStatus()` JSDoc.

## Why no dedup
The only genuinely shared surface is the `'CANCELLED'` notification trigger — two call-sites of `postCommit.run(..., 'CANCELLED')`, no duplicated logic to extract. Merging the service methods would *add* branching (fee vs no-fee, gate vs no-gate), not remove it.

## Follow-up (separate concern, not this PR)
`cancel()` records the fee/refund amounts but doesn't call Stripe to issue the refund — refunds are operator-manual today. Filing a refund-automation issue; out of scope for fee *semantics*.

## Test plan
- [x] api typecheck (0), full api unit suite (1489 pass), import boundaries OK
- [x] biome clean, lint:size (no new warns), lint:modules OK
- [x] guard test proven to fail under a fee-attaching mutation, then revert -> green

### DIFF
--- a/packages/api/src/services/booking-lifecycle.ts
+++ b/packages/api/src/services/booking-lifecycle.ts
@@ -202,6 +202,14 @@ export class BookingLifecycleService {
     )
   }
 
+  /**
+   * Operator status transition (route-gated to MANAGEMENT_READ_ROLES, #643).
+   * Transitioning to CANCELLED is the operator cancel path and is deliberately
+   * FEE-FREE — operator non-delivery is the operator's fault, so the renter is made
+   * whole, never penalised for it. The tiered cancellation fee lives ONLY on renter
+   * self-cancel (`cancel()`); the two CANCELLED paths are intentionally divergent,
+   * not duplicated (#679) — do not merge them.
+   */
   async updateStatus(
     ctx: CallerContext,
     bookingId: string,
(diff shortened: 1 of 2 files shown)


jackli921:
Refund-automation follow-up filed as #851 (the money-movement gap is out of scope here — this PR is fee *semantics* + the guard test).

Issue #887, opened 2026-06-16
Outbound email silently fails on CF Workers: global fetch called detached from globalThis (Illegal invocation)


jackli921:
**Priority: HIGH, beta-critical.** Renter confirmations *and* operator alerts are both dead in prod. Bookings still succeed (no data loss), but every email since beta launch has failed silently.

## Symptom
Booking succeeds, operator sees it in-app (red notification dot works), but **0 emails** to anyone (incl. spam). Reproduced live: booking `f4d3fc18-66e2-44af-8009-e093d2317867` — `keeponflowing2022@gmail.com` booked a KIX / Best Car Rental car; owner `li.jack0707@gmail.com`.

## Evidence (prod Neon `hidden-heart-30840607`)
`SELECT * FROM notification_log ORDER BY 1 DESC` → two rows, both `status=FAILED`, `providerMessageId=null`, and:

> `error = "Illegal invocation: function called with incorrect 'this' reference"`

Kinds `RENTER_BOOKING_CONFIRM` (to renter) + `OPERATOR_BOOKING_ALERT` (to owner) — recipients are **correct**, so this is purely the send throwing at the runtime.

## Root cause
On Cloudflare Workers, the global `fetch` is a branded builtin that must be called with `this === globalThis`. Three services capture it detached and call it as an instance method:

```ts
private readonly fetchFn: typeof fetch = fetch,   // captured unbound
// later:  await this.fetchFn(ENDPOINT, {...})     // this = the instance -> Illegal invocation
```

Node/vitest don't brand `fetch`, so CI stayed green; the first real Workers send exposed it. (Email was wired-but-untested before beta.)

## Scope — 3 files, identical bug (re-verify line numbers vs latest mp)
- `packages/api/src/services/email/resend-email-sender.ts:25` — confirmed failing in prod
- `packages/api/src/services/google-translation-provider.ts:25` — latent, same pattern
- `packages/api/src/services/geocoding/nominatim-geocoder.ts:29` — latent, same pattern

## The fix
`= fetch` -> `= fetch.bind(globalThis)` in all three constructors. Keeps the injectable test mock intact (only the default binds). Add `as typeof fetch` if TS narrows the bound type.

## Design decisions (for the implementer)
[...]

PR #892, opened 2026-06-16
fix(web)+ci: bind proxy fetch + cross-surface detached-fetch guard (#887, follows #890)

jackli921:
## Summary
Follow-up to **#890** (merged — bound the 3 API services + behavioral repro tests, the actual email-outage fix). This PR adds the **remaining runtime surface + a recurrence guard** for #887.

**Root cause recap.** On CF Workers/Pages the global `fetch` is a branded builtin: it throws `Illegal invocation` when called with `this !== globalThis`. A param/field defaulted to bare `= fetch` captures it detached.

## What's in this PR (the delta on top of #890)
1. **`packages/web/functions/_shared/proxy.ts`** — bind the Pages-Function proxy's `fetchImpl` default. It calls `fetchImpl(...)` bare (`this`=undefined), so it works today, but binding makes the invariant uniform across runtime surfaces.
2. **Cross-surface recurrence guard** — `scripts/lint-fetch-binding.ts` (pure `findDetachedFetchDefaults` + `scanRoots`) bans detached defaults across **every deployed surface**: `api/src`, `web/src`, `web/functions`, `shared/src` (`.ts/.tsx/.mts`, prod code only). Catches `= fetch`, `= globalThis.fetch`, `= self.fetch`; strips comments + strings (line-preserving) to avoid false positives. 15 TDD'd tests.
3. **CI wiring** — root `lint:fetch-binding` step **and** `bun test scripts/` (so the guard's own test — plus 7 pre-existing root lint tests — are finally enforced; `--filter '*'` never ran them).
4. **Consolidation** — removes `packages/api/tests/services/no-detached-fetch.test.ts` (the api-only guard from #890); this guard supersedes it. #890's **behavioral reproduction tests are kept**.

## Review
code-reviewer + architect reviewed the original combined change: runtime fix correct & surgical; all findings were guard-hardening, all applied here (globalThis/self variant, comment/string stripping, .tsx/.mts breadth, scanRoots coverage, CI test wiring). Deferred → **#893** (shared `boundFetch` primitive).

## Verification
- [x] `bun run lint:fetch-binding` clean (all 4 sites bound); fails on the pattern
- [x] `bun test scripts/` 108 pass; api 1524 / web 909; typecheck (api+web) + `bun run lint` clean
- [x] Merged latest mp in (resolved 3 services to #890's canonical bind — zero diff vs mp there)

## Note on #887 / live smoke
**#890 + #874 are merged**, so the outage fix and the deploy smoke-URL fix are both on trunk. #887 stays open until the owner-driven live smoke: `workflow_dispatch` deploy API worker → re-book a KIX/Best Car Rental car → both `notification_log` rows flip to `SENT`.

Part of #887.


jackli921:
Heads-up + coordination from the #890 session:

**#890 is merged** to `marketplace-pivot` (squash `8faf5c8c`) — the 3 API services are now bound (`fetch.bind(globalThis)`) with per-service runtime reproduction tests, so the actual email outage fix is on trunk. I've just kicked off a **Deploy** on mp to ship it to prod.

This PR now reads as the clean **hardening delta** on top of that, which is the right shape:
- binds the 4th site (`web/functions/_shared/proxy.ts`) for uniformity — it was never broken (bare `fetchImpl(...)`, `this=undefined`, which Workers accepts), but binding it is harmless and consistent;
- replaces my narrower api-only `no-detached-fetch` guard with the **cross-surface `lint:fetch-binding`** gate (api + web + web/functions + shared) — strictly broader, good supersession;
- wires `bun test scripts/` so the lint's own tests actually run in CI.

No concerns from my side — LGTM pending CI/review. The runtime behavioral tests from #890 already prove the bind works; this locks it in repo-wide. Thanks for catching the web/functions surface.

Issue #885, opened 2026-06-16
feat(web): redesign search results map ↔ list interaction (car-first, Option B) — UX rebuild of #458


jackli921:
## Summary
Rebuild the renter search-results map↔list interaction. Today the map view's per-vehicle cards and per-location pins are out of sync: clicking a card does nothing, "在地图上显示" only toggles a faint highlight, and clicking a pin highlights all co-located cards with no popup/info. **Design decided: Option B (car-first).** Decision brief: `docs/plans/2026-06-15-search-map-list-redesign.md` (branch `feat/search-map-redesign`, draft PR linked below). UX rebuild of #458 (original map+flat-list build, closed).

## Decision (Option B — car-first)
- Result granularity = **individual cars + class-combos**, each tied to its **pickup-store pin** — a foreign tourist instantly sees *where in Japan* a car is.
- Co-located cars share a pin; clicking it opens an **Airbnb-style carousel popup** (industry standard).
- Card focus → map **flies to + popup** on its pickup pin; bidirectional hover/selection sync.
- Price-labeled pins; geo-context (nearest landmark/station · distance · prefecture); mobile Map toggle.

## Fixes (current bugs)
- Card click does nothing → card focus flies map + opens popup.
- "在地图上显示" does nothing → replaced by a visible map response.
- Pin highlights all cards, no info → carousel popup of exactly the cars at that store.

## Suggested slices (independent vertical PRs)
1. Card↔pin sync + fly-to + pin popup (single-car case).
2. Co-location **carousel** + price pins.
3. **Geo-context** labels + mobile Map toggle.

## Built forward-compatible for
- One-way rentals #882 (pickup-labeled pins, multi-point map adapter, dropoff-open DTO, fee-term pricing).
- "Search this area" #883 (Option C).

## Open questions for sign-off (brief §5)
1. Card/popup → detail page vs inline quick-view (lean: link to detail).
2. Keep the 门店/地图 toggle, or unify into one map+list default (lean: unify).
Plus two leaning calls: price-labeled pins, mobile Map toggle.

## Next
Colleague sign-off on §5 → implementation plan → TDD build (slices above).

Issue #499, opened 2026-06-09
Vite search: preserve class + pickupLocationId filter state across search→detail→form


jackli921:
Follow-up from #497 (epic #378). The storefront search loaders accept a repeatable `class` filter and `pickupLocationId`, but no navigation re-emits them: `StorefrontSearchForm` navigate, `StorefrontCard` Link, and the detail 'back to search' link all send only `{from,to}`, and the detail→search redirect on an invalid range drops `pickupLocationId`. Today this is latent (no class-filter UI sets it), but a future filter UI would silently lose state on the first click. Thread the active filters through every navigate/Link/redirect, or formalize as 'reset on form submit'. Faithful to the Next original, so non-blocking.

jackli921:
Merged via PR #562 (chore/quick-wins → marketplace-pivot).

PR #898, opened 2026-06-16
feat(web): search rows link to the pickup-store detail via an explicit CTA (#885 slice 1b)

jackli921:
## Summary

Slice 1b of the search results **map ↔ list redesign** (#885, Option B car-first). Each search result row's previously **inert disabled "Select" button** is now an explicit **"View cars" CTA `<Link>`** to the pickup store's detail page (`/$locale/storefronts/$locationId`), carrying the active **date range + class/region/pickup filters** so the drill-down preserves search context. This is the "explicit detail CTA" half of the brief's §4/§5 *card hover/focus drives the map, only the CTA navigates* rule.

> **Stacked on #895** (slice 1). Base is `feat/search-map-s1-flyto-popup`; GitHub will retarget to `marketplace-pivot` once #895 merges.

## What changed

- `SearchResultRow.tsx` — the disabled "Select" button → a `<Link>` to `/$locale/storefronts/$locationId` with `{from, to, ...carryForwardFilters({class, pickupLocationId, region})}` (the exact helper + target `StorefrontCard` already uses, #499). Label reuses the existing `search.viewStore` key ("View cars") — the target lists the store's cars; no per-car detail route exists today.
- Threaded the search context **route → `SearchMap` → `SearchMapList` → `SearchResultRow`**, mirroring the existing `StoreGrid`/`StorefrontCard` drill.
- Removed the now-orphaned `search.select` i18n key in all three locales (en/ja/zh, lockstep).

## Scope note — card-as-affordance deferred to slice 2

Slice 1b was originally scoped as *CTA + card-as-affordance (hover/focus drives the map)*. I built the hover/focus half, but **code review caught a HIGH**: slice 1's recenter works by *remounting* `<PigeonMap>` (selectedId is in the remount key), so binding selection to `onMouseEnter` remounts the map + re-fetches tiles on **every row the cursor crosses**. The hover trigger needs to pair with an **imperative, non-remounting recenter** (the animated fly-to), which is slice 2's work. So I reverted that half and **moved card-as-affordance to slice 2**, where the fly-to mechanism lands. Slice 1 selection (marker click + "show on map" button) is unchanged here.

## Test plan

- [x] `bun run --filter @kuruma/web test` — **922/922 pass** (CTA target, filter carry-forward, and SearchMapList threading-guard tests added; i18n parity green after the key removal).
- [x] `bun run --filter @kuruma/web typecheck` — 0 errors.
- [x] `bunx biome check` — clean.
- [x] code-reviewer pass: SHIP-WITH-FIXES → all fixes applied (HIGH = defer hover; MEDIUM = delete orphaned keys).

Refs #885 (does not close — card-as-affordance + carousel + price pins + geo-context + mobile remain in slices 2/3).


jackli921:
## Pre-merge review: **SHIP** ✅

Reviewed the authoritative net diff (`git diff origin/marketplace-pivot...HEAD`, 10 files +250/−29) ahead of the owner merge.

**Critical check — the reverted hover-select → PigeonMap remount/tile-refetch did NOT return:**
- `SearchMapList` selection machinery (`selectedId`/`setSelectedId`) is unchanged — drives only the `ring-2` highlight + the existing show-on-map button. `mapItems` memo stays keyed on `items`, not selection, so a select never rebuilds the plotted pins.
- `SearchMap.tsx` +32 is pure prop-threading (6 readonly ctx props declared → destructured → forwarded to `SearchMapList`). No recenter/`setCenterZoom`, no remount path.
- Row is a plain `<article>` with no `onClick`/hover/focus; inert `<button disabled>` → explicit `<Link>` "View cars" → `storefronts/$locationId` carrying `from/to` + `carryForwardFilters(...)`. Semantic anchor, matches the brief's CTA-only rule.

**Also verified:** route search keys (`from/to/class/pickupLocationId/region`) match the target's `validateSearch` so nothing is silently stripped; `carryForwardFilters` omits empties (exactOptionalPropertyTypes-safe); i18n `search.select` dropped from all 3 locales with no dangling refs, `viewStore` present in all 3; new tests pin the exact CTA href + filter carry-forward (mutation-resistant).

**LOW (comment-only, non-blocking):** `SearchResultRow.tsx:27` docstring says "card focus/hover drives the map" — that's slice-2 behavior, not present in this diff. Suggest softening to future tense ("…will drive the map in #885 slice 2").

Clear to merge once CI (`test-and-build`, `e2e-real-db`) goes green.

Issue #882, opened 2026-06-16
feat(marketplace): one-way rentals — different pickup & dropoff locations (design needed)


jackli921:
> **Design:** [2026-06-21-one-way-rentals-design.md](https://github.com/jackli921/kuruma-rental/blob/docs/882-one-way-design/docs/plans/2026-06-21-one-way-rentals-design.md) (PR #1021) · **Scope:** same-operator one-way only; cross-operator returns are out of scope (see scope-decision comment below).

## Execution — sub-issues (atomic, dependency-ordered)
- [ ] #1022 — **Pure core**: `locationAt` + `assessPickupFeasibility` + `findStrandedSuccessors` _(no deps — dispatchable now)_
- [ ] #1023 — **effectiveEndAt follows the dropoff turnaround** (F1) _(no deps — dispatchable now)_
- [ ] #1024 — **Booking creation + cancellation continuity**, advisory-locked _(← #1022, #1023)_
- [ ] #1025 — **Search/availability parity** across all renter surfaces _(← #1022)_
- [ ] #1026 — **Web**: return-to-different-location toggle + badge + map route _(← #1025)_
- [ ] #1027 — **Drop fee** (optional, post-MVP) _(← #1024)_

v1 ships **tail-only** one-way (a one-way dropoff only when the car has no later booking); the design seam dials up to full-chain / operator-relocation / cross-operator **additively**. The model: a vehicle's location is a pure fold over its booking chain (never a mutable column). Maintainability findings F1–F7 (review 2026-06-21) are folded into the issues above.

---

## Summary
Allow a renter to pick up a car at one store and drop it off at another (one-way / relocation rental). **Surfaced during the search map↔list redesign** (`docs/plans/2026-06-15-search-map-list-redesign.md` §6).

## Why
Foreign tourists often travel city-to-city (land at KIX, tour Kyoto, fly out elsewhere). One-way pickup/dropoff is a common car-rental expectation and a real conversion lever.

## Two parts, very different cost
1. **UI/UX (cheaper).** Search toggle + dropoff field, fee badge, map dropoff/route — built forward-compatible by the map redesign (§6 guardrails).
2. **Inventory / fleet model (the real cost).** A vehicle now *ends* at a different location → availability becomes per-location-over-time. The design resolves this as a booking-chain fold + a tail-only feasibility seam. Backend epic (#1022–#1025).

## Open product decisions (design §8)
[...]

Issue #840, opened 2026-06-14
feat(marketplace): renter region search — nearest-first rank, distance labels, map-centering, nav threading (#651 Slice 3 incr 5)


jackli921:
Fast-follow to #651 **Slice 3 increment 5**, split out of the Slice 3 PR (which shipped increments 1-4: the region picker wired into live search, results filtered to a region subtree). This issue is the *polish on top* — the slice already works without it.

## Scope (one branch, one PR — TDD)
1. **Nearest-first rank** within the filtered subtree. Anchor = the chosen region's center (`latitude/longitude`, now on `StorefrontCard` — incr 1) **or** the "Near me" point. Null-coord stores sort last. **No region chosen → keep the existing `operator → name → id` sort unchanged.** Sort lives in `services/storefront-search.ts` `compareCards` (api) — thread the anchor coord through `GET /storefronts/search` (it already takes `regionId`; resolve the region's center server-side, or pass the anchor lat/lng).
2. **Distance labels** "~N km · Area" on `StorefrontCard.tsx` (web), using `haversineKm` from `@kuruma/shared/lib/region-distance`.
3. **Map-centering** on the chosen region's lat/lng (`vite/search/PigeonMapAdapter.tsx` — currently fit-all-pins via `computeViewport`; fall back to it when region coords are null).
4. **Nav threading** of the `region` slug through search → detail → back (the #499 `carryForwardFilters` pattern; `carryForwardFilters` already carries `region`, the detail route needs to accept + re-emit it).

## Verify
- TDD per design §9 web cases + service sort cases.
- **e2e real-db lane:** operator store in Kansai → renter picks an area → store appears nearest-first → book. Needs a disposable pg.

## Design source of truth
`docs/plans/2026-06-12-renter-location-search-niconico.md` §6 (precedence: a selected region always wins over geolocation; near-me is opt-in; denial → full list) and §9 (web cases). URL contract is the **slug** (Decision 6).

jackli921:
Done — Slices A (ranking) + B (map-centering) + C (nav threading) + the real-DB §9 e2e all shipped in PR #870 (squash-merged to marketplace-pivot). Client-side per page (design §6/§7); server-side ?sort=distance remains the documented #439 follow-up. web 891/891, e2e-real-db green on CI.

PR #912, opened 2026-06-17
feat(web): search map↔list slice 2 — non-remount recenter, price pins, carousel (#885)

jackli921:
## What & why

Slice 2 of the search map↔list redesign (#885), per the signed-off plan
`docs/plans/2026-06-16-search-map-slice2-plan.md`. Slice 1 (#895) added fly-to + a
single-car popup; 1b (#898) added the "View cars" detail CTA. But the fly-to
**remounted `<PigeonMap>`** (selectedId lived in the remount `key`), re-fetching
every tile on each row interaction — which is why card-as-affordance hover was
reverted out of 1b. Slice 2 makes the map genuinely interactive without the
tile-thrash.

## The 4 tasks (dependency order)

1. **Non-remount recenter** (`c11caf86`) — viewport (`center`/`zoom`) is now
   controlled state, synced from `onBoundsChanged` (user pan/zoom sticks) and
   retargeted by **one** precedence effect (`focusViewport`: selected pin → region
   anchor → fit-all). The remount `key` is gone, so a fly-to animates in place
   (pigeon `setCenterZoomTarget`) with zero tile re-fetch.
2. **Card hover/focus** (`4095caa5`) — hovering or keyboard-focusing a geocoded row
   flies the map to its pin. Idempotent select (sets, never clears) so focus→click
   can't race to a deselect; "Show on map" is now an idempotent action (no
   `aria-pressed` toggle), keeping the touch affordance for Slice 3.
3. **Price-labeled pins** (`1257b386`, `36fd995f`, `b2ec1d9d`) — additive
   `renderPin` seam on the adapter; the view owns the whole pill (group min price,
   `onClick`, styling, i18n). "From ¥X" for a multi-car store, "¥X" for one, "Price
   on request" otherwise. Selected pin inverts + `aria-current`. **P2 a11y:** each
   pill's accessible name is store + price (`map.pinSelect`) so visually identical
   prices stay distinguishable to screen readers.
4. **Co-location carousel popup** (`21adb72e`) — clicking a pin opens
   `MapPopupCarousel` over the **whole** `groupByLocation` group (mirrors
   `PhotoGallery`'s index/modulo cycling): photo · name · class · price · "View
   cars" CTA per slide, wrap-around next/prev, "n / total" position label. Keyed to
   the location so a new selection opens at the first car.

## Notable decisions

- **No remount `key`** is the crux — verified against pigeon-maps v0.22.1 source
  (controlled props animate via `setCenterZoomTarget`; `onBoundsChanged` is
  debounced 60ms so syncing it back doesn't truncate a programmatic fly-to). Reasoning in the plan.
- **Position label instead of PhotoGallery's dot row** in the carousel — reads
  cleaner in a narrow (`w-56`) popup and is the same affordance a screen reader needs.
- **One-way (#882) seam deferred** (YAGNI) — the adapter already takes a points
  list + selectedId; a dropoff pin / route is an additive prop when #882 lands.
- **Adapter contract stays clean (D1):** `SearchMapList` imports no map library;
  `renderPin`/`renderSelected` are pure `(item, state) => ReactNode` props, and the
  adapter only positions the returned nodes. `renderPin` falls back to the default
  `<Marker>` dot when absent, so the bare-adapter tests are intact.

## Test plan

- [x] `bun run --filter @kuruma/web test` — **990 pass / 150 files** (search group:
  19 SearchMapList, 4 MapPopupCarousel, 18 PigeonMapAdapter, 10 result).
- [x] `bun run --filter @kuruma/web typecheck` — clean (prod call-sites; tests aren't typechecked).
- [x] `biome check` — clean; pre-commit (file-size, module-boundaries, tsc web+api) green.
- [x] code-reviewer agent — one MEDIUM found (carousel index leaked across pin
  switches) → fixed with a `key` + regression test, folded into `21adb72e`.
- [ ] **Owner:** real-db Playwright (`region-search.auth.spec.ts`, #840 anchor) via CI.
- [ ] **Owner:** manual `bun run dev` — hover a row flies the map (no tile flash),
  pins show prices, click a multi-car pin → carousel.

Part of #885. **Owner-gated — please do not `--admin` merge.**

### DIFF
--- a/docs/plans/2026-06-16-search-map-slice2-plan.md
+++ b/docs/plans/2026-06-16-search-map-slice2-plan.md
@@ -1,0 +1,137 @@
+# Search map↔list redesign — Slice 2 (#885)
+
+## Context
+
+The search-results map is the renter funnel's core. Slice 1 (#895) added fly-to + a single-car
+pin popup; Slice 1b (#898) added an explicit "View cars" detail CTA. But the **fly-to remounts
+`<PigeonMap>`** (`selectedId` is in the remount `key`), so any per-row map interaction re-fetches
+all tiles — which is exactly why card-as-affordance hover/focus was built then **reverted** out of
+1b (code-review HIGH). Slice 2 (brief §7 + handoff) makes the map genuinely interactive:
+
+1. **Non-remount recenter** (controlled fly-to) — fly without remounting (kills the tile-thrash). *Foundation.*
+2. **Re-add card hover/focus** — now safe; reuse the reverted tests at reflog `17f9a8f6`.
+3. **Price-labeled pins** — Airbnb/Zillow "¥8,000" pills, selected inverts (replaces grey dots).
+4. **Co-location carousel popup** — N cars at one store-pin → a swipeable popup (Turo model).
+
+Signed-off design: `docs/plans/2026-06-15-search-map-list-redesign.md` §4–5 (Option B, car-first).
+
+## The crux: pigeon-maps non-remount fly-to (verified against v0.22.1 source)
+
+`<Map>` supports **controlled `center`/`zoom`** props. On a prop change its `componentDidUpdate`
+(`index.esm.js:1154`) calls `setCenterZoomTarget` → an **animated tween** (no remount). `onBoundsChanged`
+reports the post-move viewport — and `syncToProps` (which fires it) is **debounced 60ms** (`:1095`,
+`DEBOUNCE_DELAY:111`), while `componentDidUpdate` compares against the *animation target* mid-flight
+(`:1185`) and prop-equality-guards (`:1181`). So syncing `onBoundsChanged → state` does **not** truncate
+a programmatic fly-to (the debounce lets the tween settle before the prop re-syncs). Mechanism: hold
+`{center, zoom}` in state, sync from `onBoundsChanged`, push the selected pin's viewport on selection
+change → animated fly-to, no remount.
+
+**Drop the remount `key` entirely.** Re-fit on a new result set / region is handled by a controlled
+reset effect (below), which *animates* to the new fit (pigeon caps far jumps via `animateMaxScreens`,
+so a cross-country new search snaps instead of crawling). No `key` → the map never remounts → zero
+tile-thrash on any interaction.
+
+## Approach — 4 TDD tasks (dependency order), branch off `feat/search-map-s1b-card-cta`
+
+### Task 1 — Controlled-mode recenter (the no-remount foundation)
+- `PigeonMapAdapter.tsx`: replace `defaultCenter`/`defaultZoom` + `key` with controlled `center`/`zoom`
+  from `useState(() => focusViewport(pins, anchor, selectedId))`; wire `onBoundsChanged → setView`;
+  **remove the `key` prop entirely.** A **single** viewport-sync effect (one state → one precedence
+  function — this is the fix for the two-effect race): on a `targetSignature` of `selectedId` + `anchor`
+  + each pin's `id:lat:lng` → `setView(focusViewport(pins, anchor, selectedId))`. `focusViewport` already
+  encodes the precedence (**selected pin wins → else region anchor → else fit-all**), so a result-set
+  change with a *still-valid* selection stays flown-to that pin, while a *stale* selection (gone from the
+  new set) falls through to the new region/fit — no two effects can disagree. The signature includes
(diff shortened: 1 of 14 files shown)


jackli921:
## Pre-merge review (loop drain pass): **SHIP-WITH-FIXES** — the slice does its job ✅

Reviewed the net diff via the code-reviewer agent. **The make-or-break item — recenter-on-select WITHOUT remount — is implemented correctly:**
- The remount `key` on `<PigeonMap>` is **deleted** (`PigeonMapAdapter.tsx`); recenter goes through pigeon-maps **controlled** `center`/`zoom` driven by `useState` + the pure `focusViewport`, no remount/tile-refetch.
- Pinned by a **node-identity** test (`PigeonMapAdapter.test.tsx`: `expect(after).toBe(before)`) — stronger than a mount-spy; proves React reused the exact DOM node across selection.
- `mapItems` stays `useMemo`-keyed on `items`, not selection.

**Also verified:** i18n parity for all 6 new keys (en/ja/zh); carousel prev/next are real labelled buttons (keyboard-operable, chevrons `aria-hidden`); price pins are real `<button>`s with store+price aria-labels; carousel index-leak fixed (keyed on `locationId`); TS `exactOptionalPropertyTypes` clean; file sizes ≤181 (800 cap); no DB access in web; tests mutation-resistant (exact `data-center`/`data-zoom`, 4-case selection precedence).

**Non-blocking nits (all optional):**
- MEDIUM — `targetSignature` rebuilt every render via `pins.map().join()` on the pan hot-path; YAGNI at ~10-20 pins, memoize only if pin count grows.
- MEDIUM — pin "From {price}" keys off group size not price spread (a 2-car same-price location still says "From"); prefix only when `min<max` if undesired.
- LOW ×3 — `aria-current="true"`→`"location"` on pin/row; carousel card could `aria-live="polite"` for SR announce on advance; dead `?? [item]` fallbacks.

Leaving the merge to your session (slice-2 of your active #885 thread) — flagging that it's reviewed + green so you can merge when ready.

jackli921:
## Architect review (follow-up to the loop-drain pass) — **architecturally sound, 1 fix applied**

Ran an independent architect-review lens (modular design / seam extensibility / state model / types), and dispositioned the 5 nits from the pre-merge review. Net: **one a11y fix taken, four reasoned no-change/defer.**

**Fixed now** (`cef17d3f`):
- **Carousel `aria-live`** — wrapped the popup card's text region (name·class·price) in `aria-live="polite"`, scoped to the text (not the photo) so the image swap doesn't double-announce. Next/prev now announces the new car instead of swapping silently. +1 regression test (live region updates on advance). web **991 pass**, typecheck 0, biome + pre-commit green.

**No change — reasoned:**
- **"From {price}" by group size, not price spread** — *correct UX, not a bug.* Count-based means "multiple cars here, cheapest is ¥X" — surfaces inventory/choice. Spread-based would render two-¥8,000-cars identically to one-¥8,000-car, hiding the extra inventory. Matches Airbnb/Zillow convention.
- **`aria-current="true"` → `"location"`** — *keep `"true"`.* The `"location"` token is for nav position (breadcrumb/wizard), not geography despite the name; `"true"` is correct for selection-in-a-set and matches the existing `CancellationPolicy.tsx` convention.
- **`?? [item]` fallbacks** — *unreachable but intentional.* `groupItemsById` is keyed over the full `items`; the plotted item is a subset, so the key always hits. The `??` is the type-safe narrowing for `Map.get(): T | undefined` (avoids a `!`), and becomes load-bearing if the group source ever diverges from `mapItems`. Leaving it.

**Deferred (YAGNI):**
- **`targetSignature` rebuilt per render on the pan hot-path** — real in principle, but at 40-50 vehicles → ~5-15 location pins it's a sub-µs string join. Memoizing would also require memoizing `pins` (rebuilt each render) → two memos to save a 15-element join = net negative on readability. Revisit only if pin count grows by an order of magnitude.

**Seam check (forward-fit):** `MapAdapterProps` (`renderPin`/`renderSelected` render-props) extends to Slice 3 (geo-context labels = presentation inside the existing closures; mobile bottom-sheet reuses `MapPopupCarousel`) and #882 (one-way = additive `secondaryPins?`/`route?`) **without a breaking contract change.** The `key`-on-`locationId` carousel reset and the controlled-viewport `onBoundsChanged`↔`targetSignature` round-trip were both verified non-drifting (node-identity + pan-then-reset tests).

Re-pushed; CI re-running. Still owner-gated.

Issue #904, opened 2026-06-17
feat(web): operator self-service staff invites + membership management


jackli921:
Part of #902 (operator cold-start).

## Problem
An operator owner cannot invite or manage their own staff. Provider invites are **PLATFORM_ADMIN-only** today (`POST /admin/provider-invites`, gated by `requirePlatformAdmin` in `packages/api/src/routes/admin.ts`). There is no operator-facing team/members page. A real business onboarding their staff must go through a platform admin every time.

## Scope
- **API:** operator-scoped invite endpoint (e.g. `POST /operators/me/invites`) that mints an `OPERATOR_STAFF` (and maybe `OPERATOR_OWNER`) invite **only for the caller's own operatorId**, reusing `ProviderInviteService` (`packages/api/src/services/operator-grant.ts` / provider-invite service). List + revoke own pending invites; list current `operator_memberships` for the operator; deactivate a member (status → non-ACTIVE) respecting the single-active-membership invariant. Owner-only (not staff) for invite/revoke/deactivate.
- **Web:** `/$locale/_business/manage/team.tsx` (or `/settings/team`) page + nav entry + i18n. Show members (role, status), pending invites with copyable link, invite form, revoke/deactivate actions.
- **Tests:** API — operator A cannot invite into operator B; staff cannot invite; single-active-membership enforced; revoke flips status. Web — list/invite/revoke render + call. E2E: owner invites a staff email, accepts via `/provider/invite/:token`, lands scoped to the same operator.

## DoD
An operator owner invites and manages their own staff entirely through the portal without platform-admin involvement, with all writes scoped to their operator and covered by tests.

jackli921:
Starting work this session. Basing off `origin/develop` (note: `marketplace-pivot` was renamed to `develop`, now default — plan doc says mp but that branch is gone). Worktree `kuruma-904-team`, branch `feat/904-operator-staff-invites`. Following the 2-slice plan in docs/plans/2026-06-16-issue-904-operator-staff-invites.md.

jackli921:
## Refined plan (verified against `develop` tip `c898e293`)

Following docs/plans/2026-06-16-issue-904-operator-staff-invites.md with these deltas after verifying every reuse reference:

1. **Base = `origin/develop`** (doc says `marketplace-pivot`; that branch was renamed to `develop`, now default).
[...]

PR #938, opened 2026-06-17
feat(#904): operator owner revokes invites + deactivates staff (slice 2)

jackli921:
Part of #904 (operator self-service team management). Slice 2: the owner-only **write** side of the team page — revoke a pending invite, deactivate an active member. Slice 1 (read-only `/manage/team` + invite) shipped in #927.

## What's in this slice

**Shared**
- `REVOKED` added to the `provider_invite_status` enum (mig `0064`).

**API** (`/operators/me/*`, cookie + CSRF, tenant derived from session — the client never names an operatorId)
- `POST /invites/:id/revoke` — owner-only; 404 on unknown/foreign id.
- `POST /members/:id/deactivate` — owner-only; **409 on the last owner**; clears the `users` operator projection *before* revoking the membership ledger (fail-safe order).
- One **pending** invite per `(operatorId, email)` — partial unique index (mig `0065`); duplicate create → `409 ConflictError`. Revoke frees the slot.
- `markAccepted` guarded with `status='PENDING'` (race guard).

**Web**
- `RevokeInviteDialog` + `DeactivateMemberDialog` — mirror `ArchiveLocationDialog`: nullable-entity prop controls open, synchronous in-flight ref guard, mutation reset on close, server 404/409 message rendered inline.
- `TeamView` gains `canManage` + per-row buttons; each labels itself with the row subject (email / name) for a distinct accessible name.
- Route wires selected-entity state and mounts both dialogs behind the owner gate.
- en/ja/zh keys added (parity check: 1028 keys × 3).

## Deliberate scope cut (owner-approved)

Deactivation clears the membership ledger + `users` projection only. **Immediate JWT/session revocation is deferred** to a follow-up: a renter-door sign-in mints the JWT role straight from the `users` projection and `verifyAndMap` never re-checks the DB per request, so the live token can linger up to its ≤7-day TTL. Clearing the projection is necessary but not sufficient for instant cut-off. Follow-up issue filed.

## Verification
- api **1624** · web **1032** · shared **583** — all green
- web typecheck clean · api `lint:boundaries` OK · biome clean (pre-commit)
- `db:verify` 5/5 (66 migrations, journal `when` monotonic)
- i18n parity 1028 keys × en/ja/zh

## Test plan
- [ ] As operator OWNER: revoke a pending invite → row disappears, link dies.
- [ ] As OWNER: deactivate a staff member → loses portal access.
- [ ] As OWNER: attempt to deactivate the last owner → inline 409 refusal, dialog stays open.
- [ ] As OPERATOR_STAFF: no revoke/deactivate buttons render.


jackli921:
Deferred-revocation follow-up tracked in #939.

Issue #1025, opened 2026-06-22
feat(#882): one-way search/availability parity across all renter surfaces


jackli921:
Part of **#882**. **Design:** [§5.4](https://github.com/jackli921/kuruma-rental/blob/docs/882-one-way-design/docs/plans/2026-06-21-one-way-rentals-design.md). Folds **Policy Drift, parity-scope, F3, F4, F7**.

## What
Every renter availability surface must answer the **same** question creation answers — *"bookable for this exact pickup→dropoff request?"* — by **calling the pure predicate (#1022)**, not a re-encoded SQL subset.

### Repo prefilter (`packages/api/src/repositories/drizzle/availability.ts`)
- `findAvailableVehicles` keeps the road-legal gate + `NOT EXISTS` overlap and adds a LATERAL keeping only cars **physically at the pickup** at `requestedStart` (movement history, `status <> 'CANCELLED'`):
```sql
LEFT JOIN LATERAL ( SELECT b."dropoffLocationId" FROM bookings b
  WHERE b."assignedVehicleId" = v.id AND b.status <> 'CANCELLED'
    AND b."effectiveEndAt" <= :requestedStart
  ORDER BY b."effectiveEndAt" DESC LIMIT 1 ) last_leg ON true
WHERE COALESCE(last_leg."dropoffLocationId", v."pickupLocationId") = :pickupLocationId
```
- New index `idx_bookings_assignedVehicle_effectiveEnd` on `bookings(assignedVehicleId, effectiveEndAt)` (custom migration; `db:verify` green).

### Service authoritative (parity)
- `flat-search.ts` batch-loads candidates' **bracketing** legs (one `assignedVehicleId IN (...)` query — **F4: never load full history**, COMPLETED is unbounded) and runs `assessPickupFeasibility(pickup, dropoff, window, policy)` per candidate — the **identical** pure fn the booking tx uses. Bookable ⇔ creatable.
- Extend `AvailabilityRepository.findAvailableVehicles` (`packages/api/src/repositories/types.ts:497`) to carry the requested **dropoff** + **effectiveEnd**. Update **both** implementations.

### All surfaces (not just flat search)
- `flat-search.ts`, `storefront-search.ts`, `storefront-detail.ts`, and per-vehicle `checkVehicleAvailability` (`routes/availability.ts`, `GET /availability/:id` — gains the dropoff) all run the predicate.
[...]

Issue #1026, opened 2026-06-22
feat(#882): one-way web surface — return-to-different-location toggle + badge + map route


jackli921:
Part of **#882**. **Design:** [§7 slice 4](https://github.com/jackli921/kuruma-rental/blob/docs/882-one-way-design/docs/plans/2026-06-21-one-way-rentals-design.md) + search-map-redesign §6.

## What
The renter-facing one-way UI, modeled on Hertz/Avis/Turo (industry standard — support, not invention, per renter-UI guidance).
- Search form: a **"Return to a different location"** toggle (default off = same location) revealing a **dropoff** field.
- Results: a one-way badge ("One-way OK", + drop fee if #1024's follow-up ships); cars filtered to the pickup→dropoff route (the API does this — #1025).
- Map (`PigeonMapAdapter`): the pin is the **pickup** pin; optionally a dropoff pin / route line on focus. The §6 forward-compat guardrails (multi-point adapter, dropoff-open DTO) make this additive.
- i18n: en/ja/zh for toggle, dropoff field, badge.

## Acceptance criteria
- Toggling reveals/hides the dropoff field; off = same-location (today's behavior).
- A one-way search passes pickup + dropoff to the API and renders only route-valid cars.
- Component/e2e tests; don't regress region-centering (#840) or `region-search.auth.spec.ts`.

## Dependencies
**#1025** (API parity provides one-way-bookable results + the dropoff param).

## Out of scope
Backend (#1022–#1025), drop-fee pricing.

jackli921:
Deferring one-way rentals — removing from open work per owner decision (2026-06-25). Already-merged foundation pieces (#1022 pure core, #1023 dropoff turnaround) remain in code behind the `assessPickupFeasibility` seam and are harmless; nothing further is planned. Reopen if one-way is revived.

Issue #1024, opened 2026-06-22
feat(#882): one-way booking creation + cancellation continuity (advisory-locked)


jackli921:
Part of **#882**. **Design:** [§5.3 shell, §6 lifecycle](https://github.com/jackli921/kuruma-rental/blob/docs/882-one-way-design/docs/plans/2026-06-21-one-way-rentals-design.md). Folds maintainability **Check-Then-Act, F2, F5**.

## What
Wire the pure core (#1022) into the booking tx and the cancel path, serialized per vehicle.

### Booking creation (`packages/api/src/services/booking-creation.ts`)
- Inside `submitInTx` (already in an interactive tx via `runTx`), first `SELECT pg_advisory_xact_lock(hashtext(:assignedVehicleId))`. R2/tail are check-then-act across rows; the exclusion constraint only catches *time overlap*, not a *continuity* break — the lock serializes creates per vehicle.
- Load **movement legs** (`status <> 'CANCELLED'`, incl COMPLETED; **bounded** to bracketing legs), call `assessPickupFeasibility`. `ok:false` → `{ok:false, status:400, error, code}`.
- **Replace** the static check at `booking-creation.ts:236` (`vehicle.pickupLocationId !== input.pickupLocationId`) — it becomes `locationAt(legs, home, startAt) === pickup`. The same-operator dropoff guard at `:247` (the #1013 test) stays.

### Cancellation (`booking-lifecycle.ts:303`)
- Under the **same** per-vehicle lock, call `findStrandedSuccessors` (#1022).
- **F2 — renter vs operator:** a renter cancelling **within policy** (72h-free etc.) MUST NOT be blocked to protect another renter's later booking. So renter self-serve cancel proceeds and stranded successors are **cascaded to operator action** (detect + persist a stranded flag + log); the **operator** cancel path may block. Never silently leave the impossible state. (Full operator-alert UI may be a follow-up, but **detection ships here**.)

### Substitution (`booking-lifecycle.ts substitute`) — **F5**
- Take the lock on **both** vehicles (ascending id order, deadlock-safe). Validate the *gaining* vehicle via `assessPickupFeasibility`; run `findStrandedSuccessors` on the *losing* vehicle. Replace `replacement.pickupLocationId === booking.pickupLocationId` with `locationAt(replacement, startAt) === booking.pickupLocationId`.

## Acceptance criteria (real-PG integration)
- Round-trip after a one-way: pickup ≠ car's current location → 400.
[...]

PR #1028, opened 2026-06-24
fix(#1023): effectiveEndAt follows the dropoff location's turnaround (one-way)

### LINKED ISSUE #1023: fix(#882): effectiveEndAt must follow the dropoff location's turnaround (one-way)
Part of **#882**. **Design:** [one-way design §6 "Turnaround follows the dropoff"](https://github.com/jackli921/kuruma-rental/blob/docs/882-one-way-design/docs/plans/2026-06-21-one-way-rentals-design.md). Maintainability finding **F1 (High)**. **Independently shippable — no dependency on the rest of the epic.**

## Problem
`effectiveEndAt = endAt + turnaround` is set by the `compute_effective_end_at` trigger (`drizzle/0037_booking_exclusion_assigned_vehicle.sql`) from the **pickup** location:
```sql
SELECT "defaultTurnaroundMinutes" INTO _turnaround FROM locations WHERE id = NEW."pickupLocationId";
```
mirrored at `packages/api/src/services/booking-creation.ts:251`. For a **one-way** rental the car is returned and cleaned at the **dropoff**, not the pickup — so the next-bookable window at the dropoff is computed from the wrong location and the car can be advertised before it's actually been returned/cleaned. Latent today because every booking has pickup == dropoff.

## Fix
When `pickupLocationId !== dropoffLocationId`, compute turnaround from the **dropoff**:
- Custom migration updating `compute_effective_end_at()` to read the dropoff's `defaultTurnaroundMinutes` (fallback pickup/2880). `bun run db:generate --custom`; never edit merged 0037; `db:verify` green.
- `booking-creation.ts:251` reads the dropoff location's turnaround for one-way (it already loads `dropoff` at :246).

## Acceptance criteria
- Real-PG: one-way A→B with B.turnaround ≠ A.turnaround → `effectiveEndAt == endAt + B.turnaround`.
- Same-location booking unchanged; existing turnaround tests green.
- `bun run db:verify` green.

## Dependencies
None. Schema already carries `dropoffLocationId`. Can ship before the pure core.

## Out of scope
Feasibility / search.

jackli921:
Closes #1023. Part of the one-way rentals epic #882 — design §6 "Turnaround follows the dropoff", maintainability finding **F1 (High)**. Independently shippable; no dependency on the rest of the epic.

## Problem
`effectiveEndAt = endAt + turnaround` (the car's next-bookable window, enforced by the `bookings_no_overlap` exclusion constraint) was computed from the **pickup** location's `defaultTurnaroundMinutes` in two places that must agree:
- the Postgres trigger `compute_effective_end_at()` (migration `0037`), and
- its service mirror in `packages/api/src/services/booking-creation.ts`.

For a one-way rental the car is returned and cleaned at the **dropoff**, not the pickup — so the window was computed from the wrong location and the car could be advertised before it was actually back. Latent today because every booking is same-location (pickup == dropoff), where the two values are identical.

## Fix
- **`drizzle/0069_turnaround_follows_dropoff.sql`** (custom) — `CREATE OR REPLACE FUNCTION compute_effective_end_at()` keyed on `NEW."dropoffLocationId"`; recreate the trigger to fire on `(endAt, dropoffLocationId)` — the true determinant set — instead of `(endAt, pickupLocationId)`. Snapshot-invisible objects, no schema drift.
- **`booking-creation.ts`** — derive turnaround from the dropoff location, byte-identical to the trigger (`?? 2880` mirrors `COALESCE(_turnaround, 2880)`). Same-location reuses the pickup row (no extra query).
- Stale "pickup" comments on the fallback constant and the substitution path corrected (from review).

Same-location bookings are unchanged. Substitution stays invariant (it touches only `assignedVehicleId`; neither pickup nor dropoff changes). Search/feasibility are **out of scope** (#1025) — the pickup-side turnaround shown on storefront cards is left alone.

## Tests (TDD, RED→GREEN proven)
- **Real-pg integration** (`tests/integration/bookings.test.ts`): one-way A→B with B.turnaround=180 ≠ A.turnaround=120 → `effectiveEndAt == endAt + 180min`. Failed on the old pickup-keyed trigger (got 120), passes now.
- **In-memory service** (`tests/services/booking.test.ts`): one-way with dropoff=1440 ≠ pickup=2880, with a `.not.toBe(pickup)` mutation guard.

## Verification
- api integration **292/292** (real pg, throwaway `postgres:16`)
- api unit **1742/1742** (`env -u DATABASE_URL`)
- `tsc` 0 · `lint:boundaries` OK · biome clean · `db:verify` 5/5 green
- code-reviewer: **SHIP** (0 CRIT/HIGH/MED; 2 LOW stale comments fixed)

### DIFF
--- a/drizzle/0069_turnaround_follows_dropoff.sql
+++ b/drizzle/0069_turnaround_follows_dropoff.sql
@@ -1,0 +1,36 @@
+-- #1023 (one-way rentals design §6 / maintainability finding F1): "effectiveEndAt"
+-- must follow the DROPOFF location's turnaround, not the pickup's. A one-way rental
+-- is returned and cleaned at the dropoff, so the car's next-bookable window is the
+-- dropoff's buffer. 0037 keyed compute_effective_end_at() off "pickupLocationId" —
+-- correct only while every booking was same-location (pickup == dropoff), where the
+-- value is identical. This is latent until one-way bookings exist; same-location
+-- bookings are unchanged. Custom because triggers/functions are not expressible in
+-- drizzle's table builder and are snapshot-invisible (no schema-vs-snapshot drift).
+
+-- 1. Recompute "effectiveEndAt" from the DROPOFF location's turnaround.
+--    locations."defaultTurnaroundMinutes" is NOT NULL DEFAULT 2880, so COALESCE only
+--    guards an (FK-protected, effectively impossible) missing dropoff row — kept byte
+--    for byte in step with the service mirror in booking-creation.ts.
+CREATE OR REPLACE FUNCTION compute_effective_end_at() RETURNS trigger AS $$
+DECLARE
+  _turnaround int;
+BEGIN
+  SELECT "defaultTurnaroundMinutes" INTO _turnaround
+    FROM locations
+    WHERE id = NEW."dropoffLocationId";
+
+  -- ?? 2880 (48h) hard fallback if the dropoff location row is somehow absent.
+  NEW."effectiveEndAt" := NEW."endAt" + COALESCE(_turnaround, 2880) * interval '1 minute';
+  RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;--> statement-breakpoint
+
+-- 2. "effectiveEndAt" now depends on ("endAt", "dropoffLocationId"); "pickupLocationId"
+--    is no longer a determinant. Recreate the trigger over the new determinant set
+--    (CREATE OR REPLACE FUNCTION alone can't change the trigger's UPDATE OF columns).
+--    Substitution still touches only "assignedVehicleId", so "effectiveEndAt" stays
+--    invariant under substitution and the exclusion re-checks automatically.
+DROP TRIGGER IF EXISTS bookings_set_effective_end_at ON bookings;--> statement-breakpoint
+CREATE TRIGGER bookings_set_effective_end_at
+  BEFORE INSERT OR UPDATE OF "endAt", "dropoffLocationId" ON bookings
+  FOR EACH ROW EXECUTE FUNCTION compute_effective_end_at();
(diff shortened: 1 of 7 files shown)


jackli921:
## Architect review — verdict: SHIP (no blocking issues)

Second-pass review focused on the danger-zone concerns a code review doesn't cover: rolling-deploy safety, backfill, and epic-fit.

**Rolling-deploy: SAFE.** `deploy.yml` applies migrations **before** deploying the Workers, so the deploy window is *old code + new trigger*. The `BEFORE INSERT` trigger is authoritative for the stored value and `.returning()` hands it back, so every booking created in the window already gets the dropoff-derived `effectiveEndAt`. The reverse ordering never occurs.

**Backfill: not needed (verified).** Beta has 4 bookings, **0 one-way** (`pickupLocationId <> dropoffLocationId`). Production is schema-only/pre-launch. The migration only recomputes on INSERT/UPDATE of `(endAt, dropoffLocationId)` — it does not retro-fix existing rows — but there are no stale one-way rows to fix.

**Accuracy nuance.** The commit calls the bug "latent (every booking is same-location)" — true for the **web** UI (it sends `dropoff = pickup`; the one-way web surface is #1026, unshipped), but the **API already accepts same-operator one-way bookings today** (`dropoffLocationId` is a settable input; only cross-operator is rejected, per the #882 guardrail). So this fix closes a *live* correctness gap for direct/Trip.com callers, not just a future one.

**Epic-fit observation (out of scope for #1023, for #1024):** the API accepts one-way bookings before the slices that make them fully correct (search parity #1025, cancellation continuity #1024) have shipped. Worth deciding on #1024 whether to gate one-way creation behind a flag until parity lands, or treat direct-API one-way as an advanced/unsupported path meanwhile.

No code change required — scope (turnaround only) is correct; search/feasibility correctly left to #1025.

Issue #464, opened 2026-06-06
feat(marketplace): class-combo deals + inventory-count availability (post-demo)


jackli921:
From §1.1 / §5. Post-demo fast-follow. Renters book a **class deal** (exact car assigned by the operator on pickup day).

**Acceptance**
- `CLASS_COMBO` bookings (no `assigned_vehicle_id` at booking time).
- Availability via **per-(operator, location, class, time-range) inventory count** (the per-vehicle exclusion constraint cannot enforce class-only — no assigned vehicle).
- Operator assigns a concrete vehicle on/before pickup day (ties into substitution / `assigned_vehicle_id`).
- Surface class-combo cards in the map/flat-list search.

**Design:** depends on `fulfillment_mode` (#463). Open question = exact inventory-count model (§4 item 2).
Refs: epic #385; §1.1, §4, §5.

---

# Slice 2d implementation plan

> Checkpoint for the next build slice. Source: `docs/plans/2026-06-23-464-slice-2d-combo-create.md` on branch `feat/464-class-combo-rehome`. **Awaiting sign-off — no code yet.** Incorporates 4 review rounds.

Date: 2026-06-23 · Branch: `feat/464-class-combo-rehome` · Issue: #464 (§1.1/§4/§5)
Status: **PLAN — awaiting sign-off, no code yet.** Danger zone (booking tx + composition root).

## Goal
Flip the slice-2b `501 NOT_IMPLEMENTED` gate so `CLASS_COMBO` bookings actually
create. No car assigned at booking time; availability is guarded by an
**advisory lock + `countClassDemand` vs class supply**, and the price comes from
the **class rate plan** (slice 2a), not a vehicle's rates.

## Why this is a danger zone
- Adds the **first advisory lock** in the codebase (combo has no per-vehicle
  exclusion constraint to prevent oversubscription).
- Threads **two new repos into `TransactionRepos`** + the Drizzle tx factory +
  the composition root — `lint:boundaries` territory.

## Decisions needed before coding (3)
**D1 — Supply definition.** Inventory gate **rejects when `demand >= supply` ->
`409 CLASS_SOLD_OUT`** (so `demand == supply` is sold out — the next insert would be
`supply + 1`). Pin the exact boundary in a test: with `demand == supply` the create
returns 409, with `demand == supply - 1` it returns 201.
`demand = availabilityRepo.countClassDemand(op, class, location, from, to)` (2c).
[...]

PR #1077, opened 2026-06-25
feat(#464): render CLASS_COMBO cards in cross-operator search (slice 3b)

jackli921:
## What

Slice 3b of #464 (class-combo deals). The read-side producer (slice 3a, `b666a1a1`) already emits `CLASS_COMBO` search-result cards, but the web `SearchResultRow` had `default: return null` and **silently dropped every combo row**. This slice renders them.

- **`ComboRow`** (new, in `SearchResultRow.tsx`): class label as title, a **"Class deal"** badge, the live **inventory count** (`availableCount`) + seats, the daily class rate, and a CTA to the pickup store — carrying dates + active filters forward, exactly like `SpecificRow`. No transmission line (a combo has no assigned car yet).
- **`MapPopupCarousel`** made combo-aware: the badge marks the deal instead of repeating the class label (which is already the title), plus the availability count. The `aria-live` region still announces the combo count on slide change.
- The switch dropped its `default` branch — the two-member union is now **exhaustively** covered, so a future `kind` is a compile error rather than a silent null.

## Product decisions (settled before building)

- **CTA funnels to the storefront detail page** (the only detail surface today). A dedicated *book-the-class* flow is a later #464 slice.
- The producer filters `availableCount <= 0`, so a combo card always has ≥1 available — **no sold-out state needed**.

## i18n

New `search` keys in **en / ja / zh**: `classDeal`, `available` (ICU plural), `comboHint`. Parity guard green (1091 keys × 3).

## Tests (TDD)

7 new tests (6 `SearchResultRow` + 1 `MapPopupCarousel`): combo title/badge, inventory count (singular + plural), no-transmission, daily price, CTA carry-forward, and the carousel deal badge + count (asserting the label isn't duplicated).

## Verification

- `bun run --filter @kuruma/web test` — **1247 passed**
- web typecheck, i18n parity, biome, `vite build` — all green
- pre-commit hooks (biome / file-size / module-boundaries / tsc web+api) — green
- code-reviewer: no CRITICAL/HIGH; applied its test-tightening suggestion

Refs #464

### DIFF
--- a/docs/runbooks/2026-06-24-consent-signing-key.md
+++ b/docs/runbooks/2026-06-24-consent-signing-key.md
@@ -61,3 +61,23 @@ fail-closed protection:
 Once a production deploy is gated on this secret, add `CONSENT_SIGNING_KEY` to
 **`deploy.yml`**'s required-secret presence loop so a missing key fails the deploy
 loudly at the presence check instead of waiting for the first accept to 500.
+
+## Producing a consent evidence export (#877)
+
+Each acceptance is a self-contained, replayable evidence record: it snapshots the
+exact disclosed document (`documentSnapshot`) and the signed canonical version, so
+the export reproduces *what the renter actually agreed to*, independent of later
+edits to `consent_documents`.
+
+- **CLI:** `bun run consent:evidence -- <acceptanceId>` (or `--user <id>` / `--booking <id>`)
+  prints a JSON bundle: identity, the frozen document, the signature, and a
+  `verification.status`.
+- **Admin API:** `GET /admin/consent/acceptances/:id/evidence` (platform-admin only).
+- **`verification.status`** is the trust signal. Only `VERIFIED` is proof. Others are
+  honest gaps, surfaced verbatim: `SNAPSHOT_MISSING` (legacy/un-backfilled row),
+  `SNAPSHOT_HASH_MISMATCH` (snapshot text altered), `UNSIGNED`, `KEY_UNAVAILABLE`
+  (signed under a key we no longer hold — see #1050), `SIGNATURE_MISMATCH`.
+- **Backfill legacy rows:** `bun run db:backfill-consent-snapshot` snapshots a null
+  row *only when the current document is provably the signed artifact* (a stale or
+  edited document is skipped and reported, never backfilled with wrong text). Design
+  detail: `docs/superpowers/specs/2026-06-24-consent-evidence-export-design.md`.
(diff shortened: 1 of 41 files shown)


jackli921 (commented):
## Read-only review (drive-by; not the PR author)

Solid slice — TDD shape, discriminated-union narrowing, and i18n parity all look right. Three notes; only one I'd act on, and I'll file it as a follow-up PR after this lands so it doesn't gate the merge.

### MEDIUM — exhaustive switch has no runtime sink (cross-deploy hazard)

`packages/web/src/vite/search/SearchResultRow.tsx:37-42`, `packages/web/src/vite/search/MapPopupCarousel.tsx:61`

The kind-switch in `SearchResultRow` and the `current.kind === 'CLASS_COMBO' ? … : …` ternary in `MapPopupCarousel` rely purely on TS compile-time exhaustiveness. Web (CF Pages) and API (CF Workers) deploy independently here, so an unknown `kind` reaching a stale web bundle would:

- `SearchResultRow` — switch returns `undefined` → React renders nothing → row silently vanishes.
- `MapPopupCarousel` — silently falls into the `classLabel` branch (non-empty but wrong semantics).

Cheap runtime sink:

```ts
switch (item.kind) {
  case 'SPECIFIC':
    return <SpecificRow item={item} {...ctx} />
  case 'CLASS_COMBO':
    return <ComboRow item={item} {...ctx} />
  default: {
    const _exhaustive: never = item
    return null
  }
}
```

I'll file a small follow-up PR with the guard once this merges — no need to block on it.

> **Learn: Exhaustiveness Without a Runtime Sink**
> A switch with no `default` only proves coverage at the build that compiled it.
> Across a split deploy, a stale web bundle hits a new `kind` and silently renders empty/wrong — no error, no signal.
> Heuristic: every exhaustive switch over a wire-protocol union gets a `_: never` default branch.

### LOW (cosmetic) — ja/zh plural shape

`packages/web/messages/{ja,zh}.json:1198` — `{count, plural, other {残り#台}}` is CLDR-canonical (ja/zh have only `other`), so it's correct. Worth either a one-line `// CLDR: ja/zh have only 'other'` comment or collapsing to `残り{count}台` to avoid two future readers misreading it as a missing `one` branch. Not a defect.

### LOW — MapPopupCarousel test gap

`packages/web/tests/vite/search/MapPopupCarousel.test.tsx` — the new combo badge/count assertions don't verify those values live inside the `aria-live="polite"` region. A refactor that moved them outside the announced text wouldn't fail. One `userEvent.click(next)` → `findByText('3 cars available')` on a combo slide would seal it. Skippable for this slice.

— posted from a separate session; no worktree touched.

PR #1078, opened 2026-06-25
feat(#464): render CLASS_COMBO search cards (web read-side, slice 3b)

jackli921:
## What

Slice 3b of #464 — the **web read-side** for cross-operator search `CLASS_COMBO` cards. The API producer (slice 3a, `b666a1a1`) already emits combo rows; the web dropped them on the floor (`default: return null`). This renders them.

- **New `ClassComboRow`** + dispatcher `case 'CLASS_COMBO'`: class label as title, an inventory-count pill, seats only (no transmission — a class spans gearboxes), shared store/geo/price/CTA.
- **Sold-out handling:** the API emits `availableCount = max(0, cap − demand)`, so `0` is possible. A 0-count combo renders a muted **"Sold out"** pill with **no booking CTA** (no live link into an empty class), but stays visible/legible.
- **Composite React key fix:** `searchResultKey()` returned a bare `classId` for combos, so the same class offered at two storefronts collided into one React `<li key>`. Now `${locationId}:${classId}` — mirrors the API cursor `itemKey` (`c:<loc>:<class>`), minus the kind prefix the list doesn't need.
- **i18n** en/ja/zh: `comboAvailable`, `comboSoldOut`.
- **Refactor:** the two row variants duplicated thumbnail/location/geo/CTA markup → extracted `RowThumbnail`/`RowLocationLine`/`RowGeoLine`/`RowDetailCta` + a `RowCardProps<T>` alias. Behaviour-preserving.
- **Exhaustiveness guard:** dispatcher `default` now asserts `item satisfies never` — a future result `kind` added without a case becomes a **tsc error**, not a silently-dropped row (the exact bug this slice fixes for combos). Still returns `null` at runtime for a skewed API.

## Test plan

- `result.test.ts` — composite key dedups the same class across two storefronts (+ SPECIFIC unchanged).
- `SearchResultRow.test.tsx` — combo renders class label / availability / seats / price / CTA; omits transmission; **sold-out → "Sold out", no CTA**; SPECIFIC row unaffected.
- ✅ Full web suite **1248 passing**, i18n parity guard green (en/ja/zh), `tsc` clean. Reviewed by `code-reviewer` (SHIP) + `architect` (SHIP-WITH-NITS — both nits applied).

## Scope / follow-ups

`Refs #464` (not Closes) — issue stays open for the remaining slices.
- **Slice 5+** (next): operator assigns car on pickup day; rate-plan CRUD UI; e2e.
- **Noted, out of scope:** `MapPopupCarousel` renders combos via kind-agnostic accessors but doesn't yet surface availability / sold-out state — confirm intended when the popup gets dedicated combo treatment.

### DIFF
--- a/packages/web/messages/en.json
+++ b/packages/web/messages/en.json
@@ -1195,6 +1195,8 @@
     "auto": "Automatic",
     "manual": "Manual",
     "viewStore": "View cars",
+    "comboAvailable": "{count} available",
+    "comboSoldOut": "Sold out",
     "showOnMap": "Show on map",
     "list": {
       "heading": "Available cars",
--- a/packages/web/messages/ja.json
+++ b/packages/web/messages/ja.json
@@ -1195,6 +1195,8 @@
     "auto": "オートマ",
     "manual": "マニュアル",
     "viewStore": "クルマを見る",
+    "comboAvailable": "残り{count}台",
+    "comboSoldOut": "在庫なし",
     "showOnMap": "地図で表示",
     "list": {
       "heading": "空車",
--- a/packages/web/messages/zh.json
+++ b/packages/web/messages/zh.json
@@ -1195,6 +1195,8 @@
     "auto": "自动挡",
     "manual": "手动挡",
     "viewStore": "查看车辆",
+    "comboAvailable": "剩 {count} 辆",
+    "comboSoldOut": "已租罄",
     "showOnMap": "在地图上显示",
     "list": {
       "heading": "可用车辆",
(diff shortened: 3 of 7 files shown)


jackli921:
Closing in favour of #1077 (same slice 3b, opened first, already reviewed, and carries the combo-aware MapPopupCarousel this PR deferred). Avoiding duplicate/competing slice-3b cards.

Salvaging this PR's unique, non-overlapping value:
- **`searchResultKey` duplicate-React-key fix** (`result.ts`) — #1077 doesn't touch `result.ts`, so the same class at two storefronts still collides into one `<li key>` once #1077 makes combos render. Re-cut as a focused conflict-free follow-up off develop.
- **DRY row-fragment refactor + `satisfies never` exhaustiveness guard** (`SearchResultRow.tsx`) — these conflict with #1077's rewrite, so they're deferred to a follow-up on top of #1077 after it merges. Branch `feat/464-combo-web` kept (not deleted) so commits `78716d90`/`37afc684` stay recoverable.

Issue #627, opened 2026-06-13
Seed demo payment_events so the admin revenue tab shows data on real DB


jackli921:
Follow-up to #462 (admin Partner Revenue tab, merged via #625).

The tab sums successful \`payment_events\` rows, but the seed currently creates none — so on a freshly seeded real DB the tab renders its (correct) empty state until a live Stripe webhook fires.

**Ask:** seed a handful of \`payment_events\` in \`packages/shared/src/db/seed-data/seed-bookings.ts\` (FK → bookings), spread across a couple partners and JST months, with \`grossJpy / platformFeeJpy (4%) / netToPartnerJpy\` locked the same way \`lib/commission.ts\` does at webhook time. Then the demo/real-DB e2e can assert real figures.

Out of scope for #462 (read-only aggregation shipped + unit/e2e-covered via the mock track).

jackli921:
Done in #639 (squash `2d48f8f` → marketplace-pivot). Each non-cancelled demo booking now seeds a SUCCEEDED payment_event (gross=totalPrice, 4% split via computePlatformCommission), stepped ~35d/partner so the #462 revenue tab renders 3 partners across 3 JST months on a fresh DB. Verified via real-pg smoke + CI e2e-real-db. Follow-up: #628 (?month= filter).

PR #1127, opened 2026-06-25
feat(#1067 slice 1): reviews schema + repo + DI

jackli921:
## Summary

Slice 1 of #1067 (Airbnb-style double-blind mutual reviews — a P1 hard requirement). Lays the data foundation only; no submission service or reveal-on-read yet (next slice, #1082).

**Phase 1 — `@kuruma/shared`:**
- Enums: `REVIEW_AUTHOR_ROLES` / `REVIEW_SUBJECTS` / `REVIEW_MODERATION_STATUSES`.
- `db/review.ts` `reviews` table — `unique(bookingId, authorUserId, subject)` + 3 CHECKs + 5 FKs + 6 indexes (migration **0075**).
- Pure `lib/review-reveal.ts` (`decideReveal` / `summarizeSides` / `aggregateRatings`) — Functional Core, no I/O.
- `validators/review.ts` (submit / edit / report zod; `subRatings` `.strict()` optional dims).

**Phase 2 — `@kuruma/api`:**
- `Review` interface (`stores.ts`); `ReviewRepository` port (`insert` + `findByBookingId`, `NewReview` = `Omit<id|createdAt|updatedAt>`).
- `InMemoryReviewRepository` mirrors the unique seal by constraint name (payment-event pattern); `DrizzleReviewRepository` + `reviewColumns`/`toReview`.
- DI: `reviewRepo` added to `Repos` + all 3 composition branches + app-overrides + completeness-guard test. `pg-errors.ts` `REVIEWS_AUTHOR_SUBJECT_CONSTRAINT`.
- `index.ts` untouched — no `ReviewService` this slice; the `Repos` bundle is the wiring (an unused binding would trip `noUnusedLocals`).
- Migration **0076** `idx_reviews_bookingId` — `lint:fk-indexes` requires every FK column its own index; appended (not folded into already-applied 0075).

## Test Plan

- [x] api unit — 1912 passed
- [x] shared — 732 passed
- [x] reviews real-pg integration — 7 passed (3 positive + all 4 seals fire by Postgres-emitted name)
- [x] InMemory repo unit — 4 passed (TDD RED→GREEN)
- [x] typecheck (api + shared) clean; `lint:boundaries` OK
- [x] `db:verify` 5/5 (77 migrations); `lint:fk-indexes` 68/68

Closes #1081
Refs #1067

PR #1131, opened 2026-06-26
fix(e2e): scope real-db cleanup to spec bookings, not seeded demo rows

jackli921:
## Repo-wide CI break (date-triggered 2026-06-26 UTC)

From **2026-06-26 UTC**, every `e2e-real-db` run fails two specs — on **develop and every in-flight branch**, not any one PR. develop was green through 2026-06-25 (8 consecutive runs); the first failures are timestamped 2026-06-26T00:02Z.

```
PostgresError: update or delete on table "bookings" violates foreign key
constraint "payment_events_bookingId_bookings_id_fk" on table "payment_events"
  at cleanupFutureSarahBookings (e2e/real-db/*.auth.spec.ts)
```

- `marketplace-happy-path.auth.spec.ts` — body passes, **afterAll teardown** throws
- `region-search.auth.spec.ts` — `afterAll` throws **and** test-2's `KIX_STORE` heading assertion fails

## Root cause

`cleanupFutureSarahBookings` (duplicated in both specs) selects Sarah's bookings with a broad lower bound and deletes `notification_log / booking_events / threads / bookings` — but **never `payment_events`**:

```sql
WHERE u.email = 'sarah@example.test' AND b."startAt" >= '2026-07-01'
```

The seed gives **every non-cancelled demo booking a SUCCEEDED `payment_events` row** (#627/#639, admin revenue tab). Sarah's seeded `bk_demo_sakura_premium_van` is **`startOffsetDays: 5`** (now + 5d), CONFIRMED:

| Date (UTC) | now+5 | vs `>= '2026-07-01'` | cleanup |
|---|---|---|---|
| 2026-06-25 | 2026-06-30 | outside | clean |
| 2026-06-26 | **2026-07-01** | **inside** | FK-fails |

### Cascade into the region-search assertion
When marketplace teardown throws, Playwright **retries** the test; each retry books another **KIX** vehicle (the booking succeeds — only teardown fails), leaking bookings. KIX seeds **only 4 vehicles**, so the leak exhausts availability for the `2026-07-15..17` window and `region-search` test-2's "KIX heading visible" assertion fails.

## Fix

Exclude bookings that carry a `payment_events` row — the precise, **date-independent** discriminator between seeded and spec-created bookings:

```sql
WHERE u.email = ${RENTER_EMAIL}
  AND b."startAt" >= '2026-07-01'
  AND NOT EXISTS (SELECT 1 FROM payment_events pe WHERE pe."bookingId" = b.id)
```

Why not an id-prefix filter: the seed inserts bookings with **`seedId()` UUIDs**, not their `bk_demo_*` slugs, so `id NOT LIKE 'bk_demo_%'` is a no-op (the FK-blocking id in CI is a runtime UUID). Every non-cancelled seeded booking gets a SUCCEEDED `payment_events` row (#627/#639); this spec's **instant-book** bookings never do — so "has a payment" cleanly separates them, and it also sidesteps the seeded booking's other restrict-FK children. Seeded demo rows survive untouched (`demo-walkthrough`'s revenue tab only screenshots them, rendering clean even when empty).

## Test plan
- [ ] `e2e-real-db` goes green (both specs)
- [x] `biome check` clean on both specs

Discovered while landing #1128 (audit M3) — orthogonal to that diff.


jackli921:
Superseded by develop's #1127, which fixed the same date-triggered cleanup bug by deleting payment_events in cleanupFutureSarahBookings (e2e/real-db/marketplace-happy-path.auth.spec.ts:168) before the booking DELETE. This PR's NOT-EXISTS approach is no longer needed.

Issue #1104, opened 2026-06-25
tracker: architecture & maintainability audit follow-ups (2026-06-25)


jackli921:
## Tracker — architecture & maintainability audit follow-ups

Source: `docs/audits/2026-06-25-solid-architecture-audit.md` (read-only audit by 5 parallel architect agents across api / web / shared; lenses: SOLID + FC/IS + DDD-lite + deep modules).

**Verdict: architecturally strong.** SOLID's high-value outcomes are already enforced structurally (DIP via repo ports + composition root, the #713 service split for SRP, `Pick<>` tx bundles for ISP, pure `lib/` core for FC/IS, discriminated unions throughout). These follow-ups are targeted maintainability/correctness fixes — **not** "refactor for SOLID's sake."

Findings: **1 HIGH, 7 MEDIUM, 9 LOW.** Each is filed as a standalone, agent-takeable child issue below with an embedded plan for review.

### Cross-cutting themes
1. **Policy/scoping lives at the route, copy-pasted, not in the service** — DRY + a defence-in-depth gap (the 5× cross-operator read guard is the worst case). 
2. **InMemory↔Drizzle parity is wiring-tested, not behavior-tested (LSP)** — services branch on Postgres constraint names; only the key-set is asserted, never equivalence. This is what let the HIGH bug reach prod.
3. **Duplicated pipelines/tails** — booking submit tails, the two search services, per-entity write-route branches.
4. **Invariants enforced by comment, not type** — integer-yen money and `Session.user.role` cross boundaries as bare `number`/`string`.
5. **SSoT drift** — `messaging.ts` hides indexes in migrations; the web module-barrel lint is a no-op.

### Deferred / accepted-convention (tracked here, not filed as action slices)
- **L8 — money as unbranded `number`** (`lib/commission|cancellation-policy|pricing`): the integer-yen invariant is comment-enforced. Audit verdict: accepted convention, don't churn wholesale; introduce a `Jpy` brand at the `lib/` money signatures only *if* we decide the guard is worth it. Revisit, don't action blindly.
- **L9 — `consent-canonical.ts` imports `node:crypto`**: latent only (consent signing is API-only today); breaks edge import if the renter UI ever verifies a `contentHash`. Swap to `globalThis.crypto.subtle` when web needs it; until then a one-line "API-only subpath" note suffices.

### Child issues

**P1 — fix next (the real bug + the test gap that let it through):**
[...]

Issue #1124, opened 2026-06-25
refactor(api): enforce cross-operator read scope in the service layer, not 5 copy-pasted route guards (audit M3)


jackli921:
From the 2026-06-25 SOLID/architecture audit (M3 — high-ROI, security-adjacent).

## Problem
The identical "bypass-scope caller (PLATFORM_ADMIN / legacy STAFF/ADMIN) must name an operatorId or pass includeAll=true, else 400" block is **copy-pasted in 5 read routes**: `fee-schedules`, `insurance-options`, `add-ons`, `locations`, `notifications`. The services are anemic pass-throughs (`return this.repo.findAll(ctx, filters)`), and the repo defaults an `all`-scope caller with no `operatorId` to **every operator's rows**. So that route `if`-block is the only thing standing between a bypass caller and a cross-tenant config leak.

Verified: only these 5 routes call the services' `findAll`; internal consumers (booking-creation) call the repo directly with SYSTEM_CONTEXT + explicit operatorId, so they're unaffected.

*Heuristic (audit theme 1): if deleting one route's `if`-block would leak data, the invariant belongs below the route.*

## Fix (safe-by-default)
- `tenancy.ts`: generic pure `applyCrossOperatorReadScope<F extends { operatorId?: string }>(ctx, read, filters)` — passes filters through for operator/renter scope; for `all` scope throws `ScopeRequiredError` unless `read.operatorId`/`read.includeAll`; stamps `operatorId` when given.
- New `ScopeRequiredError` mapped to **400** in `error-handlers.ts` (mirrors `OperatorRequiredError`→422).
- 5 services' `findAll(ctx, read, filters?)` call the helper; 5 routes drop the duplicated block and pass `read = { operatorId: query, includeAll: query==='true' }`.

A new route that calls `service.findAll` now physically cannot omit the scope decision → the leak gap is closed, not just DRY'd.

## Tests
- tenancy unit: operator/renter pass-through; `all` + neither → throws; `all` + operatorId → stamped; `all` + includeAll → unfiltered.
- existing route tests stay green (behaviour identical: 400 / scoped / cross-operator).

Refs the audit. Siblings L4 (write-route schema-pick dup) + M4 (POST /bookings actor rewrite) are separate follow-ups.

PR #1134, opened 2026-06-26
refactor(#1107): push cross-operator read-scope below the route + dedupe scoped-create (audit M3+L4)

### LINKED ISSUE #1107: refactor(api): push cross-operator read-scope into services + extract scoped-create helper
**From:** architecture audit `docs/audits/2026-06-25-solid-architecture-audit.md` (findings M3 + L4). Tracker: #1104.

## Problem
**M3 — Cross-operator read-scope guard lives only at the route, copy-pasted 5x.**
The identical "bypass role + no `operatorId`/`includeAll` → 400" block is the *only* thing stopping a PLATFORM_ADMIN/PARTNER from reading every operator's private config (the services are pass-throughs; the repo defaults a bypass caller to *all* tenants). Any new caller that forgets the guard silently leaks cross-tenant data; a policy change is a 5-file edit.

Locations:
- `packages/api/src/routes/fee-schedules.ts:58-65`
- `packages/api/src/routes/insurance-options.ts:51-58`
- `packages/api/src/routes/add-ons.ts:51-58`
- `packages/api/src/routes/locations.ts:42-49`
- `packages/api/src/routes/notifications.ts:38-45`

**L4 — Write routes duplicate the bypass-vs-operator schema selection + operatorId resolution.**
Each `POST` repeats the same schema-pick + `resolveWriteOperatorId`. The planned class-rate-plan CRUD will copy it a 5th time.

Locations:
- `packages/api/src/routes/fee-schedules.ts:87-101`
- `packages/api/src/routes/insurance-options.ts:81-95`
- `packages/api/src/routes/add-ons.ts:81-95`
- `packages/api/src/routes/locations.ts:72-86`

**Pattern:** DRY + Defence-in-Depth Gap (read) / divergent-copy write-route branch.
**Heuristic:** If removing one route's `if`-block would leak data, the invariant belongs in the service.

## Approach (for review)
This is a security-adjacent change across 5 config routes — owner should pick the read-side approach before an agent starts.

**Read side (M3):**
- **Option A (recommended):** Push the "bypass caller must scope explicitly" default *into each service's `findAll`*, so the safe default is independent of the caller. A bypass role with no explicit scope is rejected (or returns empty) at the service layer; the route `if` becomes redundant defence-in-depth rather than the sole gate. Strongest defence-in-depth: a forgotten route guard can no longer leak.
- **Option B:** Extract `applyCrossOperatorReadScope(c, ctx, filters)` into `tenancy.ts` and call it from all 5 routes. Removes the copy-paste but the invariant still lives at the route layer (a new caller that skips the helper still leaks).

Recommended default: **Option A** (closes the defence-in-depth gap, not just the DRY one).

**Read-side contract — must preserve the intentional `includeAll=true` admin path.** Today `includeAll` is *only* a route query flag; it never reaches `filters`, so a service that simply rejects bypass callers with no `operatorId` would also reject the legitimate platform-wide read. Carry the intent explicitly below the route:
- `operatorId` → narrows to one tenant (as today).
- `?includeAll=true` → becomes an explicit `includeAllOperators: true` on the filter object, set **only** by the route.
- A bypass caller with **neither** is rejected at the service layer.

This keeps the audit's goal intact: deleting a route guard must not leak data, while explicit platform-wide reads still work. Add `includeAllOperators?: boolean` to each `*Filters` type (`InsuranceOptionFilters`, `FeeScheduleFilters`, `LocationFilters`, `AddOnFilters`, `NotificationLogFilters`) and honour it in the repo scope resolution — a bypass caller resolves to "all" *only* when the flag is set; absent it, the safe default is reject/empty regardless of caller.

**Write side (L4):** Extract `parseScopedCreate(c, ctx, { operatorSchema, adminSchema }, resolveWriteOperatorId) -> { data, operatorId }` (a route-level helper) and call it from all 4 `POST` handlers. This is a pure DRY consolidation regardless of which read-side option is chosen.

## Plan
1. Add `includeAllOperators?: boolean` to the 5 `*Filters` types and thread it through repo scope resolution (bypass caller → "all" only when set; otherwise reject/empty).
2. Implement the chosen read-side option (A or B). Each route maps `?includeAll=true` → `filters.includeAllOperators = true` and `?operatorId=` → `filters.operatorId`, then calls the service; migrate all 5 list handlers.
3. Extract `parseScopedCreate` and migrate the 4 config `POST` handlers.
4. Keep the route `if`-blocks only if Option B is chosen; under Option A they become redundant and can be dropped (or left as belt-and-suspenders — note which in the PR).

## Driving test (TDD)
Two service/repo-level cases (not just the route) — the second guards against the `includeAll` regression:
1. **Reject:** a bypass-role caller (PLATFORM_ADMIN) with NO `operatorId` and NO `includeAllOperators` MUST NOT receive another operator's config rows. Today only the route `if` stops this, so calling the service directly with a bypass identity and empty filters returns all tenants' rows before the fix and is rejected/empty after.
2. **Preserve admin all-operators read:** the same bypass caller WITH `includeAllOperators: true` still receives rows across ≥2 operators (proves the explicit platform-wide path survives Option A).

## Acceptance
- A bypass-role caller with no explicit scope cannot read another operator's config rows — asserted at the service layer (mutation-resistant: assert the returned set is empty / rejected, not just that the route returned 400).
- **The explicit `?includeAll=true` admin path still works:** a bypass caller with `includeAllOperators: true` receives all-operator rows (the platform-wide read is preserved, not collateral damage).
- All 5 list routes and 4 `POST` routes route through the shared seam(s); no per-route copy of the scope guard or the schema-pick/`resolveWriteOperatorId` block remains (grep-asserted in the PR description).
- Existing per-operator (scoped) reads and writes still return exactly their own operator's data.

Refs #1104

jackli921:
## What

Push the cross-operator read-scope default **below the route** (audit **M3**) and dedupe the write-side scoped-create boilerplate (audit **L4**). Closes #1107. Refs #1104.

Before this change, a bypass/all-scope `findAll` (PLATFORM_ADMIN / SYSTEM_CONTEXT / public-RENTER-for-location) relied on each route remembering to scope its filter. Five read routes hand-rolled that guard; four write routes hand-rolled the same "pick operatorId from caller, never the body" create scaffold. The protection was correct but copy-pasted — one missed route = a cross-tenant leak.

## How

**M3 — read-side fail-closed default.** A new `includeAllOperators?: boolean` lands on the 5 `*Filters` and a backstop in all 10 `findAll` implementations:
- **InMemory**: predicate excludes every operator unless `includeAllOperators === true` or `filters.operatorId` is set.
- **Drizzle**: `sql\`false\`` short-circuits to zero rows under the same condition.

Routes thread the flag explicitly from `?includeAll=true`. The default is now **fail-closed**: a bypass scope reads *nothing* unless the route opts in. `booking-creation.ts` (`:391`/`:658`) passes `operatorId`, so booking flows are unaffected.

**L4 — write-side dedupe.** `parseScopedCreate` (in `routes/helpers.ts`) centralizes "resolve operatorId from the authenticated caller, never trust `body.operatorId`" for the 4 create routes (add-ons, fee-schedules, insurance-options, locations).

**Blast-radius fix:** the in-memory storefront repo now passes `includeAllOperators: true` — the public catalog is an *explicit* cross-operator read, mirroring the Drizzle storefront. ~8 test files that relied on the old implicit-leak behavior were updated to the explicit contract.

## Contract

> A bypass/all-scope `findAll` returns rows **only** when `filters.includeAllOperators === true` or `filters.operatorId` is set. Default = zero rows. The operator role never reads `body.operatorId`.

## Verification

- `api` **1936 unit + 329 integration** green on real Postgres
- `tsc --noEmit` clean, `lint:boundaries` OK, biome clean
- **code-reviewer + architect both SHIP**, no must-fix (confirmed: fail-closed default, route-only opt-in, `sql\`false\`` zero-rows, never trusts `body.operatorId` for operator role, no weakened tests)

## Deferred follow-ups (architect, all SHIP-optional)

- Extract `resolveOperatorFilter(ctx, filters)` to collapse the 10-fold inline precedence (InMemory/Drizzle divergence risk).
- `parseReadScope` read-guard twin — 5 routes still dup the 400.
- `SYSTEM_CONTEXT = all` footgun: future internal `findAll` callers silently get `[]` (current 2 callers are safe).

## Review note

Security-sensitive (tenancy isolation) — intended for human review/merge, not auto-merge. The two agent reviews above already passed.

### DIFF
--- a/packages/api/src/repositories/drizzle/add-on.ts
+++ b/packages/api/src/repositories/drizzle/add-on.ts
@@ -23,6 +23,11 @@ export class DrizzleAddOnRepository implements AddOnRepository {
       conditions.push(eq(addOnOptions.operatorId, scope.operatorId))
     } else if (filters?.operatorId) {
       conditions.push(eq(addOnOptions.operatorId, filters.operatorId))
+    } else if (!filters?.includeAllOperators) {
+      // scope is 'all' (bypass role) without an explicit platform-wide opt-in:
+      // read nothing. The safe default lives here, so a forgotten route guard
+      // cannot enumerate every tenant's private config (#1107).
+      conditions.push(sql`false`)
     }
 
     if (filters?.status) {
--- a/packages/api/src/repositories/drizzle/fee-schedule.ts
+++ b/packages/api/src/repositories/drizzle/fee-schedule.ts
@@ -21,6 +21,11 @@ export class DrizzleFeeScheduleRepository implements FeeScheduleRepository {
       conditions.push(eq(feeSchedules.operatorId, scope.operatorId))
     } else if (filters?.operatorId) {
       conditions.push(eq(feeSchedules.operatorId, filters.operatorId))
+    } else if (!filters?.includeAllOperators) {
+      // scope is 'all' (bypass role) without an explicit platform-wide opt-in:
+      // read nothing. The safe default lives here, so a forgotten route guard
+      // cannot enumerate every tenant's private config (#1107).
+      conditions.push(sql`false`)
     }
 
     if (filters?.status) {
--- a/packages/api/src/repositories/drizzle/insurance-option.ts
+++ b/packages/api/src/repositories/drizzle/insurance-option.ts
@@ -26,6 +26,11 @@ export class DrizzleInsuranceOptionRepository implements InsuranceOptionReposito
       conditions.push(eq(insuranceOptions.operatorId, scope.operatorId))
     } else if (filters?.operatorId) {
       conditions.push(eq(insuranceOptions.operatorId, filters.operatorId))
+    } else if (!filters?.includeAllOperators) {
+      // scope is 'all' (bypass role) without an explicit platform-wide opt-in:
+      // read nothing. The safe default lives here, so a forgotten route guard
+      // cannot enumerate every tenant's private config (#1107).
+      conditions.push(sql`false`)
     }
 
     if (filters?.status) {
--- a/packages/api/src/repositories/drizzle/location.ts
+++ b/packages/api/src/repositories/drizzle/location.ts
@@ -21,6 +21,11 @@ export class DrizzleLocationRepository implements LocationRepository {
       conditions.push(eq(locations.operatorId, scope.operatorId))
     } else if (filters?.operatorId) {
       conditions.push(eq(locations.operatorId, filters.operatorId))
+    } else if (!filters?.includeAllOperators) {
+      // scope is 'all' (bypass role) without an explicit platform-wide opt-in:
+      // read nothing. The safe default lives here, so a forgotten route guard
+      // cannot enumerate every tenant's private config (#1107).
+      conditions.push(sql`false`)
     }
 
     if (filters?.status) {
--- a/packages/api/src/repositories/drizzle/notification-log.ts
+++ b/packages/api/src/repositories/drizzle/notification-log.ts
@@ -150,6 +150,12 @@ export class DrizzleNotificationLogRepository implements NotificationLogReposito
     const conditions: SQL[] = []
     if (scope.kind === 'operator') conditions.push(eq(notificationLog.operatorId, scope.operatorId))
     else if (filters?.operatorId) conditions.push(eq(notificationLog.operatorId, filters.operatorId))
+    else if (!filters?.includeAllOperators) {
+      // scope is 'all' (bypass role) without an explicit platform-wide opt-in:
+      // read nothing. The safe default lives here, so a forgotten route guard
+      // cannot enumerate every tenant's private ledger (#1107).
+      conditions.push(sql`false`)
+    }
     if (filters?.bookingId) conditions.push(eq(notificationLog.bookingId, filters.bookingId))
 
     const rows = await this.db
--- a/packages/api/src/repositories/in-memory/add-on.read-scope.test.ts
+++ b/packages/api/src/repositories/in-memory/add-on.read-scope.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from 'vitest'
+import type { CallerContext } from '../../middleware/auth'
+import type { AddOn } from '../../stores'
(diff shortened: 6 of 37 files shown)


jackli921:
Parked as draft pending #1128 (issue #1124). Both PRs fix audit M3 on the same 5 read routes from different layers; per swarm coordination #1128 (service-layer scope-or-400) lands first. This PR will then rebase to keep only its unique value — the L4 write-side `parseScopedCreate` + a thin data-layer fail-closed backstop — and drop the overlapping read-route edits. (Also: current build is red only via a merge-artifact — `types.ts` 803 > 800 cap once combined with develop's #1127/#1130; resolved by that rebase.)

jackli921:
ℹ️ **#1128 merged the service-layer M3 approach** to develop — this PR's repo-layer cross-operator read-scope is now **redundant and conflicting** (branch is DIRTY; build red on \`lint:size\` — \`repositories/types.ts\` 803>800).

Its unique remaining value is the **L4 scoped-create dedup** (not covered by #1128). Recommend: **close this and re-file L4 as a small PR atop #1128**, or rebase dropping the now-redundant M3 changes.

Flagging for the owning session — not touching the branch.

jackli921:
Superseded — closing per the owner-decided split of #1107 into its two audit halves:

- **M3** (push cross-operator read-scope below the route) shipped as **#1128** (squash-merged to `develop`) using the `applyCrossOperatorReadScope` + `ScopeRequiredError` approach. This PR's alternative fail-closed-default approach to the same guard is not the one we kept.
- **L4** (dedupe operator-scoped create) is redone cleanly in **#1160**, scoped to just the `parseScopedCreate` helper + the 4 config routes + a focused security test.

Nothing unique is lost: the M3 surface is already on `develop` via #1128, and the L4 surface lands via #1160. Closing to avoid the entangled M3+L4 diff and its conflicts with develop.

PR #1157, opened 2026-06-26
refactor(#1107): dedupe write-side scoped-create into parseScopedCreate (audit L4)

### LINKED ISSUE #1107: refactor(api): push cross-operator read-scope into services + extract scoped-create helper

jackli921:
## What
Re-files the **L4 write-side dedup** from #1134 cleanly atop `develop`.

#1134 mixed L4 (write) with M3 (read-side scope), but **#1128 already shipped M3** in the service layer (`tenancy.ts`). So #1134's M3 changes are now redundant + conflicting, and that PR is a stale DRAFT. This PR carries only L4 — the unique value #1128 explicitly deferred — and builds on #1128's `tenancy` module.

## How
`parseScopedCreate` in `routes/helpers.ts` centralizes the create-route branch (bypass caller parses the admin schema naming a target `operatorId`; operator caller parses the plain schema and is stamped its own tenant server-side — `body.operatorId` is never trusted for an operator role). The 4 create routes (add-ons, fee-schedules, insurance-options, locations) drop the copy-pasted boilerplate. Pure refactor — behavior identical.

## Verification
- `tsc --noEmit` clean, `biome check` clean
- new `scoped-create.test.ts` 4 passed; the 4 route suites 69 passed (73 total green)

Closes #1107. Supersedes the L4 portion of #1134 (that draft to be closed).

### DIFF
--- a/packages/api/src/routes/add-ons.ts
+++ b/packages/api/src/routes/add-ons.ts
@@ -1,4 +1,3 @@
-import type { CreateAddOnInput } from '@kuruma/shared/validators/add-on'
 import {
   createAddOnSchema,
   platformAdminCreateAddOnSchema,
@@ -15,8 +14,8 @@ import {
 import type { AddOnService } from '../services/add-on'
 import type { AddOnFilters } from '../services/filters'
 import type { AddOn } from '../stores'
-import { type ResolveWriteOperatorId, operatorReadScope } from '../tenancy'
-import { fail, ok, parseBody, parseId, stripUndefined } from './helpers'
+import type { ResolveWriteOperatorId } from '../tenancy'
+import { fail, ok, parseBody, parseId, parseScopedCreate, stripUndefined } from './helpers'
 
 export function createAddOnRoutes(
   service: AddOnService,
@@ -70,24 +69,15 @@ export function createAddOnRoutes(
       if (!FLEET_WRITE_ROLES.has(user.role)) return fail(c, 'Forbidden', 403)
 
       const ctx = toCallerContext(user)
-      // Bypass callers must name the target operator in the body; operator
-      // callers never send one — their tenant is stamped server-side. Resolve
-      // operatorId inside each branch where the body type is concrete.
-      const isBypass = operatorReadScope(ctx).kind === 'all'
-
-      let d: CreateAddOnInput
-      let operatorId: string
-      if (isBypass) {
-        const parsed = await parseBody(c, platformAdminCreateAddOnSchema)
-        if (!parsed.ok) return parsed.response
-        d = parsed.data
-        operatorId = await resolveWriteOperatorId(ctx, parsed.data.operatorId)
-      } else {
-        const parsed = await parseBody(c, createAddOnSchema)
-        if (!parsed.ok) return parsed.response
-        d = parsed.data
-        operatorId = await resolveWriteOperatorId(ctx)
-      }
+      const parsed = await parseScopedCreate(
+        c,
+        ctx,
+        { operatorSchema: createAddOnSchema, adminSchema: platformAdminCreateAddOnSchema },
+        resolveWriteOperatorId,
+      )
+      if (!parsed.ok) return parsed.response
+      const d = parsed.data
+      const operatorId = parsed.operatorId
 
       const result = await service.create(ctx, {
         operatorId,
--- a/packages/api/src/routes/fee-schedules.ts
+++ b/packages/api/src/routes/fee-schedules.ts
@@ -1,4 +1,3 @@
-import type { CreateFeeScheduleInput } from '@kuruma/shared/validators/fee-schedule'
 import {
   createFeeScheduleSchema,
   platformAdminCreateFeeScheduleSchema,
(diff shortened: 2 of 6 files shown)


jackli921:
Superseded by #1160, which merged the identical audit-L4 `parseScopedCreate` dedupe to develop (`cd44c5c4`); #1107 is now CLOSED. This was a parallel session's duplicate of the same work — same helper, same 4 config write routes; only the test-file location differed (`tests/routes/scoped-create.test.ts` in the merged version vs `src/routes/` here). Nothing unique is lost. Closing as superseded.

Issue #1109, opened 2026-06-25
refactor(api): extract shared snapshot+insert tail across booking submit modes


jackli921:
**From:** architecture audit `docs/audits/2026-06-25-solid-architecture-audit.md` (finding M1). Tracker: #1104.

## Problem
Duplicated money-path tail across the two booking submit modes. The insurance/fee/add-on snapshot, `composeBookingTotal`, walk-in mint, `bookingRepo.create`, and `BOOKING_CREATED` append are near-identical in both `submit*InTxLocked` paths; they differ only in base-price source and `fulfillmentMode`. This is the largest service in the package (712 lines, near the file-size cap). A fee fix applied to one path silently mis-prices the other — and it's money.

Locations:
- `packages/api/src/services/booking-creation.ts:373-498` (SPECIFIC mode)
- `packages/api/src/services/booking-creation.ts:642-746` (CLASS_COMBO mode)

**Pattern:** DRY / SRP — divergent-copy of pricing-snapshot logic.
**Heuristic:** When two methods differ only at the head (resolve+price) and share the tail (snapshot+insert+append), extract the tail.

## Plan
Extract a private `snapshotAndInsert(ctx, repos, { operatorId, classId, basePriceJpy, assignedVehicleId, requestedVehicleId, fulfillmentMode, effectiveEndAt, ... })` method that both `submit*InTxLocked` paths call for the shared tail (snapshot + `composeBookingTotal` + walk-in mint + `bookingRepo.create` + `BOOKING_CREATED` append). Each submit path keeps only its mode-specific price derivation and passes the derived `basePriceJpy` + `fulfillmentMode` in. No public interface change.

## Driving test (TDD)
Add an assertion that both modes produce the *same fee-snapshot shape* for an equivalent fixture: construct a SPECIFIC booking and a CLASS_COMBO booking with the same base price and inputs, and assert their fee/insurance/add-on snapshot objects are structurally identical (same keys, same computed totals). Write this RED-ish against the current duplicated paths to pin current behavior, then refactor under it. Existing SPECIFIC + CLASS_COMBO creation tests must stay green with byte-identical totals/snapshots.

## Acceptance
- Both `submit*InTxLocked` paths call a single shared `snapshotAndInsert`; the snapshot/insert/append code exists in exactly one place (grep-asserted in the PR).
- Existing SPECIFIC and CLASS_COMBO creation tests pass unchanged — identical persisted totals and fee snapshots (no behavioral drift).
[...]

Issue #1119, opened 2026-06-25
security: GET /bookings ungated + PARTNER shares bypassScope → possible cross-operator booking read


jackli921:
## Finding (from #1075 slice-6 reuse review, 2026-06-25)

`GET /bookings` (`packages/api/src/routes/bookings.ts:23`) has **no role gate** — only `requireUser`. Booking scope is decided by `bookingReadScope` (`packages/api/src/tenancy.ts:54-55`), which returns `{ kind: 'all' }` whenever `ctx.bypassScope` is set. `bypassScope` covers **both** PLATFORM_ADMIN **and PARTNER** (`packages/shared/src/auth/roles.ts:92` — `SCOPE_BYPASS_ROLES = roleSet('PARTNER', 'PLATFORM_ADMIN')`), and `scopeConditions()` returns `[]` for `{kind:'all'}` (`repositories/drizzle/booking.ts:16-22`) → unrestricted DB read.

**Consequence:** a 3rd-party PARTNER API key (e.g. Trip.com) can call `GET /bookings` and read **every operator's bookings**, cross-tenant. Compare `GET /dashboard/overview` (`routes/overview.ts:24`) which calls `requireManagementRead(ctx)` for exactly this reason ("RENTER/PARTNER → 403").

## Question for owner

Is cross-operator PARTNER read **intended** (Trip.com as a platform-wide channel), or should a PARTNER see only **its own** bookings (the ones it created)? If the latter, this is a tenant-isolation leak.

## Proposed investigation

1. Confirm the intended PARTNER booking-read scope with the product owner.
2. If scoped: add a partner-scope branch to `bookingReadScope` (PARTNER → only own-sourced bookings) instead of `{kind:'all'}`, and/or gate `GET /bookings` with `requireManagementRead`.
3. Add an integration test asserting a PARTNER key does NOT see another operator's booking.

Surfaced while scoping #1075 slice 6 (#1092), which builds a separate `requirePlatformAdmin`-gated `/admin/bookings` and does not depend on this fix.

Refs #1075 #1092

jackli921:
[...]

PR #1166, opened 2026-06-26
refactor(#1108): move booking-actor resolution into a service policy (audit M4)

### LINKED ISSUE #1108: refactor(api): move booking-actor resolution out of POST /bookings handler into a service policy
**From:** architecture audit `docs/audits/2026-06-25-solid-architecture-audit.md` (finding M4). Tracker: #1104.

## Problem
`POST /bookings` reshapes the create command by role *inside the handler*. The route computes `isManualBooker`, then overrides `renterId`/`source`/`walkInCustomer` and gates the disclaimer; the service trusts the resolved values. "Who may book on behalf of whom and force `source=MANUAL`" is domain authz — it is not reusable and not unit-testable outside Hono.

Location:
- `packages/api/src/routes/bookings.ts:157-211`

**Pattern:** SRP / Business Logic in Route Handler.
**Heuristic:** A route may *reject* on role; the moment it *rewrites the payload* on role, that's the service's job.

## Plan
Move the actor-resolution policy out of the handler and into the service layer:
1. Pass `ctx` (the auth/identity context) + the raw parsed input to `service.create` (no pre-rewriting at the route).
2. Add a `resolveBookingActor(ctx, input)` policy function in the service layer that derives `renterId`, `source`, and `walkInCustomer` from the caller's role and input — exactly the logic the route does today. Mirror the consent gate, which is already service-side, for placement and style.
3. The route shrinks to: parse input → call `service.create(ctx, input)` → map result to HTTP. It may still *reject* on role (e.g. 403) but no longer *rewrites* the payload.

## Driving test (TDD)
Unit-test `resolveBookingActor(ctx, input)` directly, WITHOUT a Hono context:
- A renter caller resolves to `renterId = caller.id`, `source` = the default (e.g. `WEB`), `walkInCustomer = undefined`.
- A manual-booker (operator/staff) caller resolves to `source = MANUAL`, `renterId`/`walkInCustomer` taken from the input's walk-in fields.

Write it RED first against the not-yet-extracted function. This is impossible to write today (the logic only exists inside the route closure), which is the point.

## Acceptance
- `resolveBookingActor` is a pure-ish policy callable with a plain `ctx` + input object — no Hono `Context` dependency — and is unit-tested for renter vs manual-booker with specific field assertions (exact `source`, `renterId`, `walkInCustomer` values, not truthiness).
- The `POST /bookings` route no longer mutates/rewrites `renterId`/`source`/`walkInCustomer` by role; it passes raw input to the service.
- Existing booking-creation behavior is unchanged end-to-end (renter self-book and manual/walk-in book both produce the same persisted command as before — assert via existing route/integration tests staying green).

Refs #1104

jackli921:
## Problem
`POST /bookings` reshaped the create command by role **inside the handler** — computing `isManualBooker`, then overriding `renterId`/`source`/`walkInCustomer`. Audit finding **M4** (SRP — Business Logic in Route Handler): "who may book on behalf of whom and force `source=MANUAL`" is domain authz, not reusable and not unit-testable outside Hono.

## Change
- **New pure policy** `services/booking-actor.ts` → `resolveBookingActor(ctx, request)` derives `{ renterId, source, walkInCustomer }` from caller role + raw request, **verbatim** from the old route logic. Callable with a plain `ctx` (no Hono `Context`).
- **Boundary type split**: `CreateBookingRequest` (raw — `renterId` optional) is what the route forwards; `CreateBookingInput` (resolved — `renterId: string`) stays the service command, so the verification gate / `submitInTx` / booking row are **untouched**. `BookingCreationService.create` resolves the actor first, then builds the normalized command every downstream read already consumes.
- **Route shrinks** to: parse → `service.create(ctx, rawInput)` → map. It still **rejects** on role (disclaimer + consent gates) but no longer **rewrites** the payload. `STAFF_ROLES` import dropped from the route.

## Tests (TDD, RED-first)
`booking-actor.test.ts` unit-tests the policy with exact field assertions: renter (forces DIRECT, ignores supplied renterId, drops walk-in), PARTNER (forces DIRECT), operator + platform-admin (keep source/renterId), walk-in passthrough + caller-id fallback.

## Verification
- Behavior unchanged end-to-end: 222 booking/route/manual-booking/operator-isolation tests green.
- Aggregate `bun run test`: shared 778 / api 2084 / web 1330.
- tsc ×3 + biome + `lint:boundaries` clean. **No migration.**

⚠️ Sibling **#1109** also refactors `booking-creation.ts` (submit snapshot+insert tail) — coordinate to avoid a conflict.

Closes #1108

### DIFF
--- a/packages/api/src/routes/bookings.ts
+++ b/packages/api/src/routes/bookings.ts
@@ -9,7 +9,6 @@ import {
 import { Hono } from 'hono'
 import {
   MANAGEMENT_READ_ROLES,
-  STAFF_ROLES,
   isOperatorRole,
   requireUser,
   toCallerContext,
@@ -149,19 +148,11 @@ export function createBookingRoutes(service: BookingService, consentGate: Consen
       const parsed = await parseBody(c, createBookingSchema)
       if (!parsed.ok) return parsed.response
 
-      // #589: STAFF and OPERATOR_* are "manual bookers" — they may book on behalf
-      // of a renter (renterId) and set source=MANUAL. An operator's renterId is
-      // authorized in the service against its OWN customer scope (renters with a
-      // prior booking with it), scope-first so it never probes the user table for
-      // an arbitrary id — preserving the #396/#475 enumeration defense
-      // (operator-user-isolation.test.ts). Everyone else books as themselves with
-      // source forced to DIRECT (no advance-booking-hours bypass via MANUAL).
-      const isManualBooker = STAFF_ROLES.has(ctx.role) || isOperatorRole(ctx.role)
-      // #589 1c: only manual bookers may register a walk-in customer inline; a
-      // renter's walkInCustomer is ignored (they book as themselves).
-      const walkInCustomer = isManualBooker ? parsed.data.walkInCustomer : undefined
-      const renterId = isManualBooker && parsed.data.renterId ? parsed.data.renterId : ctx.userId
-      const source = isManualBooker ? parsed.data.source : 'DIRECT'
+      // #1108 (audit M4): actor resolution — who may book on behalf of whom and
+      // force source=MANUAL — is domain authz and now lives in the service
+      // (resolveBookingActor, called by BookingCreationService.create). The route
+      // forwards parsed.data untouched. It may still REJECT on role below (the
+      // disclaimer + consent gates); it no longer REWRITES the payload by role.
 
       // #613: a renter self-serve booking must accept the liability disclaimer
       // (免责声明) at checkout — the IDP/license must be valid at pickup or the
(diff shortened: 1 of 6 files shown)


jackli921:
## Review: code-reviewer **SHIP** + architect **SHIP-WITH-NITS**

Ran both review passes against `origin/develop...HEAD`. No CRITICAL/HIGH/MEDIUM. Both verified the extraction is byte-identical to the prior route code; `tsc` clean, 7 new unit tests + 179 booking/route/isolation/consent tests green.

**Security (the load-bearing question): no regression, net improvement.** `resolveBookingActor` forces `renterId = ctx.userId` + `source='DIRECT'` for every non-manual caller (RENTER, PARTNER) and drops their `walkInCustomer`; only STAFF/OPERATOR keep a client-supplied `renterId`, still scope-authorized downstream (`listRenterIdsForOperator` → 403). The spoof path is closed and pinned by tests 1 & 3. Moving resolution into `create()` (below the route) makes it the single server-derived chokepoint — a future non-route caller can't bypass it.

### Coordination note for #1109 (same file family)
#1109 extracts the shared snapshot+insert tail (`submit*InTxLocked`). This PR only edits the **top** of `create()` + adds files → low conflict. **One hazard:** the real renter still arrives as the separate `renterId: string | null` param (walk-in = `null`), NOT `input.renterId`. The extracted tail must keep taking that param — if it switches to `input.renterId` trusting the "always present" resolved type, walk-ins break (booked under the operator's id). See `booking-creation.ts:478-484`.

### For owner awareness (pre-existing, not introduced here)
`resolveBookingActor` forces a PARTNER (api-key) caller's `source` to `DIRECT` (`booking-actor.ts:46`), and this PR now **cements that in a test** (`booking-actor.test.ts:37-43`). This was already the route's behavior — faithfully preserved. Worth a one-line confirm that PARTNER→DIRECT is intended given Trip.com `source=TRIP_COM` ingestion (#1119) before this test becomes the canonical contract. Not a blocker.

### Deferred LOW nits (fold into #1109, which touches these files)
1. `CreateBookingRequest` hand-duplicates both arms of `CreateBookingInput` (`booking-types.ts:50-60`) — could derive via `DistributiveOmit`. Academic at 2 modes; left out to avoid adding a distributive-conditional type for a 2-arm dedupe.
2. Walk-in strip/rebuild in `create()` (`booking-creation.ts:73-81`) is dense but correct (`exactOptionalPropertyTypes`-safe). No fix required.

**Verdict: ready for owner squash-merge.** All 4 CI checks green, MERGEABLE.

Issue #1070, opened 2026-06-25
feat: indicative multi-currency price display (JPY → renter's currency)


jackli921:
## Problem

Every price on the web surfaces is rendered only in JPY. The central formatter
`formatJpy` / `formatVehicleRate` (`packages/web/src/lib/format.ts`) emits `¥8,000`,
and the search/storefront/booking surfaces interpolate raw yen into i18n templates
(`t('fromDaily', { price: price.toLocaleString('en-US') })` in
`packages/web/src/vite/storefronts/StorefrontCard.tsx`,
`packages/web/src/vite/storefronts/AvailableVehicleCard.tsx`, and
`packages/web/src/vite/search/result.ts`).

International tourists — the core audience — reason in their home currency (USD, EUR,
AUD, CNY, etc.). Faced with a wall of `¥` figures they cannot quickly gauge whether a
car is cheap or expensive, so they bounce while browsing. The charge currency must
stay JPY (Stripe is JPY-only here, zero-decimal whole yen), but the **display** can
help renters reason about cost without changing a single yen that gets charged.

## Goal

Show an **indicative** converted price alongside the authoritative JPY everywhere
prices render — e.g. `¥27,000  ≈ $182 USD` — across:

- search result cards / map pins (`StorefrontCard`, `AvailableVehicleCard`, `search/result.ts`)
- storefront detail pages
- the reservation wizard (`reservation/AddOnsStep.tsx`, `InsuranceStep.tsx`, `ConfirmStep.tsx`)
- booking summary / confirmation (`bookings/BookingConfirmationView.tsx`)

Behaviour:

- The renter picks a **display currency**, **persisted** (localStorage), **defaulting**
  to one inferred from the active locale (en→USD, ja→JPY i.e. no conversion, zh→CNY —
  refine the mapping in design).
- FX rates are fetched **server-side**, **cached daily**, and exposed to web via a small
  read endpoint — never fetched per render.
- The converted figure always carries a clear **"indicative — charged in JPY"** disclaimer.

## Non-goals

- Charging in any non-JPY currency. The Stripe / checkout path is untouched.
- Stripe multi-currency, presentment currencies, or settlement in another currency.
- Real-time / streaming / per-request live FX. Daily cached rate is enough for an indicative figure.
- Currency hedging, rounding-to-psychological-price, or any financial guarantee about the rate.

## Proposed vertical slices

**Slice 1 — api: FX-rate provider + daily cache + read endpoint.**
[...]

PR #1209, opened 2026-06-27
test(#1070): currency follow-up nits — sr-only a11y, positive-rate seam, per-surface tests

jackli921:
The three deferred follow-up nits from the #1070 indicative-currency bug-hunt (PR #1200). Display-only — the charge is always JPY; the money path is untouched. No migration.

## What changed

1. **a11y — sr-only "approximately"** (`IndicativeNote.tsx`). The `≈` glyph reads as "almost equal to" on some screen readers and is skipped on others. Now a localized sr-only phrase ("approximately $201") is announced and the `≈` run is `aria-hidden`, so the figure is read exactly once. New i18n key `currency.approximately` (en/ja/zh). The component now uses `useTranslations`; `IntlProvider` is a strict ancestor of `CurrencyProvider` in the app shell, so this is never a new failure mode (the degradation test was retargeted to the real case: a missing `CurrencyProvider`).
2. **Positive-rate seam guard** (`currency-api.ts`). The FX network-seam schema now pins rates to `z.number().positive()`, so a `0`/negative rate fails the snapshot and the UI degrades to JPY-only instead of rendering a bogus `≈ ¥0`. `.finite()` was deliberately omitted — `z.number()` already rejects NaN/Infinity (Zod v4) and JSON can't carry them, so it'd be dead defensiveness.
3. **Per-surface wiring tests** for every surface that renders `<IndicativeNote/>` — insurance (daily price, **not** the deductible), add-on per-rental, booking total, both search-row variants, map popup, storefront + available-vehicle from-price. (The 8th surface, `ConfirmStep`, was already pinned in #1200.) A shared `tests/support/currency.tsx` harness seeds the FX query cache synchronously so each note resolves its `≈ $…` figure with no network; it deep-imports `FX_RATES_QUERY_KEY` from currency internals, which is sanctioned because it lives under `tests/` (the boundary-lint exception).

## Review + gates

- code-reviewer pass: **no CRITICAL/HIGH**. One MEDIUM was a false positive (ConfirmStep already covered by #1200); one LOW (harness localStorage hygiene) was fixed — the harness now clears the slot on entry so tests are order-independent.
- `@kuruma/web` test: **1403 passing**; web tsc clean; `lint:modules` + `lint:size` exit 0.
- Web-only change (touches `packages/web` only) — no shared/api files, so the api suite is N/A.

Refs #1070 (closed) — these are follow-ups, not a re-open.

### DIFF
--- a/packages/web/messages/en.json
+++ b/packages/web/messages/en.json
@@ -1,6 +1,7 @@
 {
   "currency": {
     "label": "Display currency",
+    "approximately": "approximately",
     "disclaimer": "Indicative only — you're charged in JPY",
     "asOf": "Rates as of {date}"
   },
--- a/packages/web/messages/ja.json
+++ b/packages/web/messages/ja.json
@@ -1,6 +1,7 @@
 {
   "currency": {
     "label": "表示通貨",
+    "approximately": "約",
     "disclaimer": "目安です — ご請求は日本円（JPY）です",
     "asOf": "レート基準日 {date}"
   },
--- a/packages/web/messages/zh.json
+++ b/packages/web/messages/zh.json
@@ -1,6 +1,7 @@
 {
   "currency": {
     "label": "显示货币",
+    "approximately": "约",
     "disclaimer": "仅供参考 — 实际以日元（JPY）结算",
     "asOf": "汇率基准日 {date}"
   },
--- a/packages/web/src/vite/currency/IndicativeNote.test.tsx
+++ b/packages/web/src/vite/currency/IndicativeNote.test.tsx
@@ -4,6 +4,7 @@ import { render, screen, waitFor } from '@testing-library/react'
 import type { ReactNode } from 'react'
 import { IntlProvider } from 'use-intl'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import en from '../../../messages/en.json'
 import { CurrencyProvider } from './CurrencyProvider'
 import { IndicativeNote } from './IndicativeNote'
 
@@ -26,7 +27,7 @@ function renderInProvider(ui: ReactNode, locale = 'en') {
   const queryClient = new QueryClient({ defaultOptions: { queries: { retry: false } } })
   return render(
     <QueryClientProvider client={queryClient}>
-      <IntlProvider locale={locale} messages={{}}>
+      <IntlProvider locale={locale} messages={en}>
         <CurrencyProvider>{ui}</CurrencyProvider>
       </IntlProvider>
     </QueryClientProvider>,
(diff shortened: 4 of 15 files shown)


jackli921:
## Review — ✅ Ship-worthy, no CRITICAL/HIGH

Independent code-reviewer pass over `git diff develop...HEAD` (web-only, display-only, no migration). Security-inert: no user input, no auth/DB, charge stays JPY. a11y, validation, and test coverage all improve.

### Verified correct
- **a11y ordering is right and pinned.** `aria-hidden` visible `≈ {indicative}` first, `sr-only` `approximately {indicative}` second → SR announces the figure once, spelled out. The "split on `≈`" guard (`split('≈')[0]` has no digit) genuinely pins the order — reversing the spans fails it. Mutation-resistant.
- **`.positive()` rejects no legitimate rate.** Confirmed against Zod 4.4.3: `z.number()` already rejects `NaN`/`Infinity` (comment accurate) and JSON can't carry them, so `0`/negative is the only JSON-reachable bad value — and no real JPY→currency multiplier is ≤ 0. Zero false rejections. The `ParseError` test asserts the specific type, not just "throws."
- **The 7 wiring tests are not tautological.** Card/row tests use a multi-day range (e.g. 30000/day × 2–3 days) and assert the *daily* `$201`, so feeding a computed total would yield `$402`/`$603` and fail. `InsuranceStep` is strongest: asserts `$201` present **and** `$670` (deductible) absent. They pin the field, not just "a note rendered."
- **No boundary violation.** `tests/support/currency.tsx` reaching `@/vite/currency/currency-api` for `FX_RATES_QUERY_KEY` is exempt (`lint-module-boundaries.ts` excludes `/tests/` + `*.test.tsx`). i18n key `currency.approximately` present in en/ja/zh (`approximately` / `約` / `约`).

### LOW (non-blocking polish — author's call)
1. **`currency-api.ts:23` — `.positive()` fails the _whole_ snapshot, not just the bad currency.** `formatIndicativePrice` already guards `rate <= 0 → null` (`indicative-price.ts:34`), so a per-currency `0` would degrade only that currency; with `.positive()` one bad rate now degrades the entire table to JPY-only. Defensible as "reject the malformed snapshot loudly," but the inline comment is imprecise: the `≈ ¥0` it cites is already unreachable downstream, and the rendered symbol would be the _target_ currency's (`$0`), never `¥`. Either trim the comment to "fail the snapshot loud, defense-in-depth (formatter also guards per-currency)," or keep `z.number()` and lean on the formatter. Make the tradeoff knowingly.
2. **`tests/support/currency.tsx:24` — `localStorage.clear()` wipes all keys, not just the one it "owns."** Comment says "own the global we touch" but it clears more than it touches. Surgical fix: drop the `clear()` and rely on `setItem(STORAGE_KEY, 'USD')` + the seeded query cache for determinism.
3. **`tests/support/currency.tsx:21` — exported `renderWithUsdIndicative` has no explicit return type.** Returns RTL `RenderResult` (JSX-component exemption doesn't apply). Trivial: `): RenderResult`.

CI: db-drift / e2e / e2e-real-db / test-and-build all green on the up-to-date HEAD. The three LOWs are polish, not blockers.

<sub>🤖 Automated review pass via Claude Code (code-reviewer agent).</sub>

PR #1227, opened 2026-06-27
refactor(#1213): block routes importing entities from ../stores + DTO-project 3 config services

### LINKED ISSUE #1164: Platform-admin dashboard: schedule-before-GA hardening (cross-slice review follow-ups)
Follow-ups from a cross-slice code review of the #1075 platform-admin dashboard (built across parallel sessions). **None are beta-blocking** — authz is sound (every `/admin/*` route is double-gated to `{PLATFORM_ADMIN}`), and the one drafted HIGH ("unbounded consent list") was **refuted** (consent is already capped at `CONSENT_ACCEPTANCE_LIST_LIMIT=200` in the repo query). The genuinely-uncapped anomaly list is already fixed in #1163. These are the remainder, for GA hardening:

### MEDIUM — structural `/admin/*` role gate (latent, not exploitable today)
`index.ts` does `app.use('/admin/*', requireAuth())` (authn only); the role gate is then applied per-handler (`requirePlatform*(ctx)`) in the #1075 slice sub-apps (revenue, overview, anomalies, bookings, consent), whereas `admin.ts` gates structurally via `app.use('/admin/*', requirePlatformAdmin)`. A future sibling route added to a slice sub-app that forgets the in-body call would authenticate but skip role authz at the route (the service re-assertion currently catches it, so latent only).
- Fix: mount a prefix role gate per admin sub-app so the gate is structural and new siblings inherit it.
- ⚠️ Check whether #1156 ("route boundary hygiene — authz off path-string") already addressed part of this before implementing.

### MEDIUM — anomaly route owns an entity→wire projection (leaky layer)
`routes/payment-anomalies.ts` imports the persistence entity `PaymentAnomaly` from `../stores` and hand-rolls `toView`; sibling slices project in the **service** and return a wire DTO. The boundary linter only blocks `../repositories/*`, not `../stores`, so it slipped CI — and 3 other routes (insurance-options/add-ons/fee-schedules) do the same.
- Fix: move `toView` into `PaymentAnomalyService`; consider extending `check-import-boundaries.ts` to also block `../stores` in routes.

### LOW — stale `STAFF/ADMIN` authz comments
Comments in `services/payment-anomaly.ts`, `services/admin-revenue.ts`, `web/vite/guards.ts`, `web/routes/$locale/_admin.tsx` still describe access as "STAFF/ADMIN/PLATFORM_ADMIN"; post-#487 it's `{PLATFORM_ADMIN}` only. Code is correct; comments mislead. One-pass sweep.

### LOW — `repositories/types.ts` at ~775/800 lines
Under cap, no anomaly bloat (correctly split into `types-payment.ts`), but the next slice touching the barrel should keep extracting before 800.

jackli921:
## What

Closes #1213. Follow-up to #1164 (PR #1212).

Two coupled changes, API-only, type-only & behavior-preserving (no migration):

1. **Linter — `check-import-boundaries.ts` Rule 1b.** Routes must not import persistence entity types from the `../stores` barrel (`stores.ts` is the types-only data-layer contract; a route receives an already-projected DTO from its service). This enforces the spirit-of-the-rule fix #1164 made by hand for `payment-anomalies`, so it can't regress.

2. **DTO-projecting config services.** The three offenders (`fee-schedules`, `add-ons`, `insurance-options`) now define a service-owned write-input type (`FeeScheduleUpdate` / `AddOnUpdate` / `InsuranceOptionUpdate`) — a `Pick` of the writable columns. `update()` takes the DTO; the `as Partial<Entity>` cast moves out of the route into the service. Routes drop the `../stores` import entirely.

The DTO is strictly tighter than the old `Partial<Entity>` cast — server-derived columns (`id`/`operatorId`/timestamps) are no longer assignable through a PATCH.

## Scope notes

- Offender set was exactly **3**, not the issue's estimated ~12: `payment-anomalies` was already clean (#1164) and `vehicle-classes` never imported `../stores`.
- The two architect LOW nits from #1164 (`/admin/*` path-prefix-bound floor; `requirePlatformMember` 500→401 fail-closed) are **deferred to a follow-up** — they touch `auth.ts`/`AGENTS.md`, a different subsystem, and the 500→401 change is a behavior change warranting its own test. Keeping this PR scoped to the linter + DTO refactor.

## Verification

- `bun run --filter @kuruma/api typecheck` — clean
- `bun run --filter @kuruma/api lint:boundaries` — Import boundaries OK
- Full API suite — **2250 passed** (199 files), incl. 25 RED-first linter unit tests (5 new for Rule 1b)
- biome check on touched files — clean

## Test plan

- [ ] CI green (tsc / boundaries / biome / full suites)
- [ ] Neon integration tests (`tests/integration/{fee-schedules,insurance-options}.test.ts`) — behavior-preserving, expected green under CI's DB


jackli921:
## Code review fix — pushed `30d6de32`

A post-push code-review pass caught a **4th offender** the original "exactly 3" claim missed, plus a blind spot in the new linter rule. Both fixed:

- **`routes/vehicle-classes.ts` was still coupled to `../stores`** via an inline type-position cast `as Partial<import('../stores').VehicleClass>`. The new Rule 1b provably could **not** see it: it sat after the import-line prefilter (the line starts with `stripUndefined(`, not `import `) and the regex required `from`. The linter printed "OK" over a live violation.
- **Rule 1b broadened + relocated**: now runs before the non-import skip and matches static (`from '../stores'`), dynamic (`await import('../stores')`), and type-position (`import('../stores').X`) forms. 2 RED-first tests added (would have returned `[]` against the old regex).
- **`vehicle-classes` graduated** to a service-owned `VehicleClassUpdate` DTO (Pick of the 11 `updateVehicleClassSchema` fields). No route now references `../stores` in any form (`grep` clean).
- **LOW folded**: dropped `status` from the 3 prior DTOs — no update validator accepts it and no `update()` reads it, so each DTO now equals exactly the writable PATCH surface its comment claims.

Verification: api tsc ✓ · `lint:boundaries` OK · full API suite **2252 passed** · biome ✓. Type-only, behavior-preserving, no migration.

PR #1233, opened 2026-06-27
fix(#1228): requirePlatformMember fails closed (401) + doc /admin/* floor invariant

jackli921:
## What

Closes #1228. The two architect LOW nits deferred from #1213 (PR #1227), originally surfaced in #1164. API-only, no migration.

1. **`requirePlatformMember` now fails closed with 401, not 500.** It called `requireUser(c)`, which throws a plain `Error` → mapped to 500 by `onError`. On the no-user path it now uses `getUser(c)` + `fail(c, 'Unauthorized', 401)` — mirroring `requireAuth`'s own no-user branch. The gate is now **self-sufficient and ordering-independent**: even if it ever runs without a preceding `requireAuth` (a future mis-wire), it denies cleanly instead of 500-ing. Unreachable today (`requireAuth` runs first), so this is hardening, not a live-bug fix.

2. **Documented the path-prefix-bound floor in AGENTS.md.** `requirePlatformMember()` only protects routes under `/admin/*`; a platform route mounted on a sibling prefix inherits no structural authz. The note makes "all platform-admin surfaces mount under `/admin/*`" an explicit, reviewable invariant.

## Why a doc note (not a lint guard)

A linter for "every router using `requirePlatform*` is mounted under `/admin/*`" needs cross-file analysis of the composition-root mount graph — disproportionate for a single mount point. The invariant is cheaply enforced by the documented convention + the existing per-handler `requirePlatform*` defense-in-depth. Revisit if platform sub-apps proliferate.

## Verification

- RED-first test: `401s (not 500) when no authenticated user is in context` — failed at 500 before the fix, passes after.
- `bun run --filter @kuruma/api typecheck` — clean
- `lint:boundaries` — OK · biome — clean
- Full API suite — **2261 passed** (200 files)

## Test plan

- [ ] CI green (tsc / boundaries / biome / full suites)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -44,6 +44,8 @@ Use them instead of manual `c.json({ success: true/false, ... })` construction.
 
 Enforced by `bun run --filter @kuruma/api lint:boundaries` (CI step) — which also flags any `new Drizzle*`/`new InMemory*` construction outside the composition root and the sanctioned `*-transaction.ts` factories.
 
+**Platform-admin surfaces MUST mount under `/admin/*` (#1164, #1228).** The structural read-floor `requirePlatformMember()` (mounted once as `app.use('/admin/*', requireAuth())` → `app.use('/admin/*', requirePlatformMember())` in `index.ts`) is **path-prefix-bound**: it only protects routes under that prefix. A platform route mounted anywhere else inherits NO structural authz and relies solely on its in-body `requirePlatform*` call — one forgotten call = an open admin surface. So: register every platform-admin router under `/admin/*`, never a sibling prefix. The per-handler `requirePlatform*` gates stay as defense-in-depth (and carry the stricter write-only `requirePlatformAdmin` distinction).
+
 ## Commands
 
 | Task | Command |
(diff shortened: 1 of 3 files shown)


jackli921:
## Independent code review — SOUND (no CRIT/HIGH/MEDIUM)

Verified at the code, not assumed:

- **Only the no-user path changed.** `requireUser` is literally `getUser` + throw-if-null, so the authenticated path stays byte-identical (`requirePlatformRead(toCallerContext(user))`); the 403 path (ForbiddenError→403) is unchanged and still proven by the existing tests. `fail(c,'Unauthorized',401)` is the *exact* envelope `requireAuth` emits at its own no-user branch — and returning a `Response` directly no longer depends on `onError` being wired to produce the status.
- **The `/admin/*` invariant holds today** (not aspirational). `requirePlatformMember` is mounted once (`index.ts:343`). All 7 `requirePlatform*`-gated route files register `/admin/...` paths only. The one trap — `services/operator.ts` is wired into both the admin and the non-admin `/operators/*` routers — checks out: `/operators/*` only reaches `scopeToCaller`/owner-gated methods; the 4 `requirePlatformAdmin` methods are reachable solely via `/admin/operators*`.
- **Blast radius: none.** Sole mount + the test file; nothing asserted 500 or try/caught the old throw.
- **Test is mutation-resistant** — genuinely exercises the no-user branch (no auth middleware ahead of it) and asserts both 401 and the error body.

LOW (no fix, matches the PR's own rationale): the invariant is convention+doc-enforced, not lint-checked — appropriate for a 7-file surface; revisit only if the admin surface grows materially.

PR #1286, opened 2026-07-01
fix(#1279): seal operatorId on location + vehicle-class repo update()

### LINKED ISSUE #1279: Seal operatorId on update() for location + vehicle-class repos (follow-up to #1271)
## Summary

Follow-up to #1271 / PR #1278. That PR sealed `operatorId` as an immutable tenant anchor on the `update()` path of the three **config** repos (add-on, insurance-option, fee-schedule). A code review of #1278 found the **same unsealed vector** in two sibling operator-scoped repos that were out of #1271's declared scope: **`location` and `vehicle-class`**.

After #1278, 4 of 6 operator-scoped Drizzle repos strip `operatorId` on update (`vehicle.ts` already did). `location` and `vehicle-class` are the last two without the seal.

## The vector (same as #1271)

Both repos' `update()` destructure only `id`/`createdAt` and spread the rest into `.set(...)`, so a payload `operatorId` would be written; the in-memory doubles pin only `id`. Tenant isolation on these edit paths therefore rests entirely on the upstream service guard + DTO omitting `operatorId` — not exploitable today, but a latent tenant-migration IDOR if a future second caller skips the guard or a DTO re-adds `operatorId`.

## Affected code

| Repo | Drizzle `update()` (no strip) | In-memory `update()` (pins only id) |
|------|-------------------------------|-------------------------------------|
| location | `packages/api/src/repositories/drizzle/location.ts:108` | `packages/api/src/repositories/in-memory/location.ts:83` |
| vehicle-class | `packages/api/src/repositories/drizzle/vehicle-class.ts:107` | `packages/api/src/repositories/in-memory/vehicle-class.ts:63-64` |

Precedent already in the tree: `packages/api/src/repositories/drizzle/vehicle.ts:218` strips `operatorId`, and the three config repos do after #1278.

## Priority note

**`vehicle-class` is the higher-priority of the two.** It is the FK *parent* of the fee-schedule composite FK `(operatorId, vehicleClassId) -> vehicleClasses(operatorId, id)` that #1271's fee-schedule seal depends on. A repointed `vehicleClasses.operatorId` would undermine that invariant from the parent side.

## Proposed fix (same Option A pattern as #1271)

- Drizzle: add `operatorId: _operatorId` to the `update()` destructure so it is dropped before `.set(...)`.
- In-memory: force `operatorId: existing.operatorId` after the `...data` spread.
- Verify no legitimate caller relies on `operatorId` in the update payload (the config-repo fix confirmed services pass DTOs that omit it; check the `location`/`vehicle-class` services + DTOs the same way).

## Tests

- In-memory repo unit test per entity asserting `update(id, { operatorId: '<other>' })` leaves `operatorId` unchanged while other fields apply (mutation-resistant, mirror #1271).
- Real-pg integration assertion in the existing `locations.test.ts` and a vehicle-class integration test (if present) driving the Drizzle `update` directly.

## Also (LOW, from the same review)

The add-on config repo has no real-pg integration file, so its Drizzle `operatorId` strip (shipped in #1278) is only proven at the in-memory layer — insurance and fee-schedule each got a real-pg test. Optional: add an add-ons real-pg test if/when an add-ons integration suite lands.

## Severity

Defense-in-depth / tech-debt. No current exploit path. `security`, `tech-debt`, `P3`. Refs #1271.

jackli921:
## What

`operatorId` is an immutable tenant anchor. Both `LocationRepository.update()` and `VehicleClassRepository.update()` let an `operatorId` present in the update payload flow through to the write. A caller reaching the repo directly (bypassing the service DTO) could silently re-home a row to another tenant -- a latent IDOR. `vehicle_class` is the FK parent of the fee-schedule composite FK, so its tenant is doubly load-bearing.

Latent, not a live regression: both update DTOs (`updateLocationSchema`, `updateVehicleClassSchema`) omit `operatorId`, and the services pass those DTOs straight through. This is defense-in-depth at the write layer, following the same pattern shipped for the config repos in #1271 (PR #1278).

## Fix

- **Drizzle** (`drizzle/location.ts`, `drizzle/vehicle-class.ts`): strip `operatorId` in the `update()` destructure (alongside the existing `id` / `createdAt` strip).
- **In-memory** (`in-memory/location.ts`, `in-memory/vehicle-class.ts`): pin `operatorId: existing.operatorId` after the `...data` spread (alongside `id` / `createdAt`).

No migration.

## Tests (RED-verified by reverting a strip)

- In-memory unit: `location` + `vehicle-class` `update()` ignore `operatorId` in the payload while still applying other fields.
- Real-pg integration (`tests/integration/locations.test.ts`, `vehicle-classes-route-fk.test.ts`): `repo.update({ operatorId: opB })` leaves the row on `opA`, verified by re-read across tenants. Both operators are seeded so a failed strip would actually migrate the FK (surfacing the bug) rather than 23503-rejecting on a missing operator.

## Verification

- api unit: 2382 passed
- real-pg integration: 16 passed (both touched files)
- `tsc` (api + web), `lint:boundaries`, biome: clean
- RED-verify: reverting the Drizzle location strip fails the new integration test (row re-homed to op_b)

Closes #1279


jackli921:
## Review complete (code-reviewer + architect agents)

**Consensus: ship as-is.** No CRITICAL/HIGH. The seal is correct at both layers in all 4 repo files, matches the canonical #1271 / vehicle.ts pattern, and the tests are mutation-resistant (RED-verified by reverting a strip; both operators seeded so a broken strip migrates the FK rather than being masked by a 23503; persistence re-read across tenants).

Addressed in this PR:
- **Comment idiom** (code-reviewer nit): reworded the 4 seal comments to the exact #1271 wording (`operatorId is an immutable tenant anchor (#NNNN): strip/pin it like id...`), plain dashes. Comment-only (`85cbf372`).

Filed as follow-ups (both reviewers agreed these are out of scope for this focused seal):
- **#1288** -- tenant-scope the `update`/`archive` WHERE on location/vehicle-class + the 3 config repos to match `vehicle.ts:226-228`. Closes the sibling "write a foreign row by id-spoof" vector, single-guarded today by the service `findById`. Also folds in the code-reviewer LOW to tighten `LocationUpdateData` (Omit -> Pick).
- **#1289** -- a `lint-tenant-anchor-immutability` scanner so repo #7 can't ship unsealed (6th hand-rolled seal; guard > helper > base-class per the review).

All 5 CI checks were green on `630d60f2`; re-running on the comment-only `85cbf372`.

PR #1305, opened 2026-07-01
feat(web): mobile UX primitives — ≥44px touch floor + scrollable dialogs (#1298)

jackli921:
Closes #1298 and #1306. Wave 1 of the mobile UX hardening epic (#1294) — lands first; siblings B/C/D/E rebase on it and extend the mobile Playwright lane.

Two shared primitives sit under the entire renter funnel, so fixing them here lifts touch targets and un-clips dialogs everywhere at once.

## What changed
- **Button touch floor** (`button.tsx`) — `pointer-coarse:min-h-11` (44px) in the shared cva base, plus `pointer-coarse:min-w-11` on the 4 square icon sizes. Lifts every size to the WCAG 2.5.5 / Apple HIG 44px minimum on touch/coarse pointers **without** changing fine-pointer desktop/operator density (which depends on the small `h-8`/`h-7` heights). Escape hatch documented for a legit sub-44px control (`pointer-coarse:min-h-0`).
- **Dialog scroll + pinned close** (`dialog.tsx`) — `DialogContent` caps at `max-h-[90dvh]` and clips (`overflow-hidden`); an **inner** `dialog-scroll` region (`min-h-0 overflow-y-auto`) scrolls the body, so a viewport-tall dialog reaches its footer instead of clipping on small phones **and** the close (X) button stays pinned instead of scrolling away with the content (#1307). Dropped 10 hand-rolled `max-h-[90vh] overflow-y-auto` overrides across the operator config dialogs + `ConsentGate`, which now inherit the mobile-correct `dvh` behaviour.
- **Off-primitive submit** (`documents/DocumentUploadCard.tsx`) — routed a raw `bg-red-600 <button>` to `<Button>` so it inherits the floor and drops the off-token colour.

## Completion gate (epic #1294)
- New **iPhone 13 (WebKit) `mobile-safari`** Playwright project scoped to `e2e/mobile/`; `chromium` scoped to desktop specs (a project-level `testIgnore` **replaces** the config-level one, so the `real-db` exclusion is repeated). CI installs `webkit` in both e2e jobs, with the browser cache key bumped so the addition isn't skipped by a stale cache hit.
- `e2e/mobile/primitives.spec.ts` proves, on a real phone:
  - **touch targets** — every Button primitive clears 44px (a genuine regression guard — without the floor these render 32px);
  - **dialog scroll + pinned close (#1306)** — a CONFIRMED booking is created through the mock, the renter `CancelBookingDialog` is opened on a short phone viewport, and the popup is asserted to cap at 90dvh, the body to genuinely overflow, the close to stay pinned outside the scroll region (RED-verified: nesting the close back inside the scroll region fails the test), and the footer submit to be reachable via scroll.
- Enabling the dialog in E2E needs `VITE_FEATURE_CANCELLATION=true` in the mock-lane web server (beta ships it OFF). Existing chromium specs never touch the confirmation page, so all 14 mock-lane specs stay green. The mock booking fixture also gained the required `addOnSnapshot` (it had drifted from `bookingDtoSchema` since the slice-6 spec was retired in #1255), plus empty `/consent/status` + `/threads` stubs so the newly-reached renter-authenticated path is deterministic.

## Test plan
- [x] `bunx playwright test` — 14 pass (12 chromium + 2 mobile-safari iPhone 13/WebKit), verified against the branch rebased on current develop.
- [x] Web `dialog`/`button`/`DocumentUploadCard` unit tests assert the exact class contracts (CI `test-and-build`); this change touches no web source, only the e2e lane.
- [x] `tsc --noEmit` (web + api), biome, `lint:size`, `lint:modules` green (pre-commit + CI).

## Review notes (folded)
`code-reviewer` + `architect` reviewed the primitives. Fixed the HIGH Playwright cache-key bug. The earlier deferral of an in-browser scroll test (it had no renter trips-list mock and reaching an admin dialog crossed bounded contexts) is resolved here by driving the renter confirmation page's own `CancelBookingDialog` — no cross-context reach.


jackli921:
Independent review pass from a parallel session — agrees this is **shippable**. Two low-severity notes, neither blocking:

**1. \`vh\`→\`dvh\` narrows the consent dialog's height cap on older Safari.**
\`ConsentGate\` dropped its local \`max-h-[90vh]\` to inherit the base \`max-h-[90dvh]\`. \`dvh\` is Safari 15.4+, so on older Safari the declaration is invalid and dropped — and the *blocking* consent dialog then has **no** height cap there (nothing local left to fall back on). Modern behaviour is strictly better; if you want to preserve the pre-2022 Safari cap, make the base defensive with both: \`max-h-[90vh] max-h-[90dvh]\` (vh fallback, dvh override).

**2. (nit) Lost \`disabled:cursor-not-allowed\` on the upload submit** — functionally equivalent, since the \`Button\` primitive uses \`disabled:pointer-events-none\` (which already suppresses the cursor). Flagging only so it reads as a conscious drop, not an oversight.

Everything else verified correct: the coarse-pointer floor (min-h/min-w win over the fixed \`h-*\`/\`size-*\` because they're different properties), the chromium \`testIgnore\` correctly repeating \`**/real-db/**\` (project-level replaces config-level), and the mobile spec's non-vacuity (the \`/en/search\` submit renders 32px without the floor, so the 44px assertion is a real guard).

PR #1360, opened 2026-07-02
fix(#1260): bind booking status writes to the picked operator (slice 2)

jackli921:
## What

`PATCH /bookings/:id/status` now binds the write to the operator the caller picked via `?operatorId=`. A `PLATFORM_ADMIN`'s `bookingReadScope` is `all`, so `findById` handed it any operator's booking by raw id and the "read-only preview" was client-side only. Now: no pick -> **422 `OPERATOR_REQUIRED`**, wrong pick -> **404** (no existence oracle), owning operator -> proceeds. Tenant operators are unaffected (already clamped by their read scope).

Reuses slice 1's `assertFleetWriteWithinOperator`; adds a shared `fleetWriteDenialResult` mapper (422/404) and widens the lifecycle result unions.

## Scope (deliberately narrow)

- **updateStatus only.** `substitute`/`assign` routes are `isOperatorRole`-gated, so a bypass admin 403s before the service — no cross-tenant hole to close there (per `docs/architecture/booking-authz.md`, platform staff deliberately excluded).
- **Web unchanged.** Both `updateBookingStatus` UI callers (`BookingActionsPanel`, `TodayPanel`) return `null` unless `isOperatorSession`, so a picker-admin never reaches the write from the UI — threading `pickedOperatorId` there would be dead code.
- Manual `POST /bookings` (slice 3) and vehicle writes (slice 4) are separate.
- **#1260 stays OPEN** — do not close on merge.

## Tests / gates (all green locally)

- Pure guard + mapper, `BookingService` facade, route binding (422/404), real-pg integration (precondition proving the hole + 422/404/200). RED-verified by neutering the guard.
- tsc x3, biome, boundaries, csrf-writes, api 2541u, web 1853u, **full integration 475** (the slice-1 lesson: ran the whole suite, no other `/status`-as-admin test regressed).


jackli921:
Code-review pass (code-reviewer agent): **no CRITICAL/HIGH.** Three LOW findings; two folded into `919303f9`:
- Dropped the dead `| 422` from `SubstituteResult` — `substitute` never produces 422 (out of scope, operator-only), so the widened union member was dead. `StatusTransitionResult`'s 422 is the real one.
- Scoped `fleetWriteDenialResult`'s docstring: the mapper is wired to booking status writes today; block/vehicle slices adopt it later.
- Skipped (acceptable): an operator-context HTTP route test — the 'guard ignores tenant operators' contract is already pinned at the service layer.

Re-verified after the fold-in: api tsc clean, api 2541u, full integration 475.

jackli921:
## Expanded to guard `/cancel` (slice 2b, `373b06f5`)

Per the review, `POST /bookings/:id/cancel` had the same (worse — refund-moving) cross-tenant hole as `/status`: no route gate, so a bypass admin could cancel any operator's booking by raw id. Now bound to the picked operator via `?operatorId=` (422 unbound / 404 non-owning / 200 owner), same contract as `/status`.

Key design point: `/cancel` admits **renters** (self-cancel), so the fleet guard (keyed `!isOperatorRole`) would wrongly 422 them. Introduced `assertBookingWriteWithinOperator` keyed on `bookingReadScope==='all'` (bypass tier only); both booking writes now share a common `operatorBindingDenial` core. `assertFleetWriteWithinOperator` (blocks/vehicles) keeps its `!isOperatorRole` key, behavior-identical.

### Re-review (2 agents)
- **Correctness**: no CRITICAL/HIGH. Refactor clean; `updateStatus` fleet→booking key migration confirmed behavior-preserving; tests mutation-resistant (incl. renter-still-cancels regression + `actor-derivation`'s 'admin can cancel any booking' converted to unbound-422/bound-200).
- **Security**: `/cancel` bypass-admin hole **CLOSED**, no new gap. One **pre-existing MEDIUM** (not a regression): a PARTNER can cancel any TRIP_COM-sourced booking cross-operator (partner scope clears the guard; `/cancel` has no route gate) — `booking-authz.md` omits `/cancel` from the PARTNER exclusion. Filed as **#1367** (needs a gate-or-sanction policy call).

### Gates (all green)
api 2554u, full integration **478** (+3 real-pg /cancel), tsc x3, boundaries, biome; CI test-and-build/db-drift/e2e/e2e-real-db all pass. `booking-authz.md` updated. **#1260 stays OPEN** (manual booking / vehicle writes remain).

PR #1350, opened 2026-07-02
feat(#1277): self-serve operator registration

jackli921:
## Summary

Self-serve operator registration end-to-end (Slices 1-4 of the plan).

- **Public application flow** — landing CTA -> `/business/register` form -> `POST /operator-applications`. Zod-validated, with a live-email partial-unique constraint (one PENDING/APPROVED application per email) surfaced as a 409.
- **Admin queue** — list + filter of applications, reject route/UI (with reason), and approve route/UI.
- **Approval provisioning** — atomic transaction: creates the operator, membership, and mints a provider invite. C1 guard blocks approval when the contact email already holds a live invite anywhere. Emits `OPERATOR_APPLICATION_APPROVED` / `OPERATOR_APPLICATION_REJECTED` audit events. The one-time invite link is revealed once in the admin UI (copyable, announced to assistive tech via `role=status`).
- **Schema** — `operator_applications` table (migration `0097`) + audit-kind enum extension (migration `0098`).
- **Refactor** — extracted `OperatorApplicationRepository` into its own module to keep the `repositories/types` barrel under the 800-line cap.

Slice 5 (email delivery) is intentionally deferred — the invite link is delivered through the admin UI for now.

## Test plan

- [x] API vitest: 2543/2543
- [x] Web vitest: 1847/1847
- [x] Typechecks: shared / api / web clean
- [x] Lints: i18n-parity (1437 keys), unwrap-schema, fk-indexes (75), file-size
- [x] Web build clean
- [x] E2E `operator-registration` passes
- [ ] Reviewer: sanity-check approval provisioning + one-time invite reveal in a running env

Closes #1277

PR #1398, opened 2026-07-03
fix(#1277 follow-ups): bound admin queue, remint owner invite, localize labels + a11y

jackli921:
Resolves review follow-ups from the self-serve operator registration work (#1277, now closed). API + web, no schema migration.

Closes #1370

> **#1371 coordination:** #1371 (admin list cap + slug-collision 409) is owned by the parallel **PR #1399**, which implements it with cleaner keyset pagination. Slice A below still carries an *earlier, overlapping* implementation of #1371 (offset pagination + slug-409); that overlap will be dropped in favor of #1399 at rebase-time once merge order is settled. The items unique to this PR's Slice A — #7 (honeypot), #9 (approve() extraction), #12 (race-fence) — stay.

## Slice A — admin queue hardening (`f1023de5`)
- **#7** Honeypot submissions get a silent `201` (validator relaxed + route trap) instead of a tell-tale 400.
- **#9** `approve()` extracted into `provisionApproval`/`provision` + module-level `assertEmailUnclaimed`.
- **#12** Approval race-fence rollback proven via a transactional in-memory harness.
- _(#1371a bounded list + #1371b slug-409 also present here — superseded by #1399, see note above.)_

## Slice B — admin re-mint of an owner invite (`965e8caa` API + web, #1370)
- `POST /admin/operator-applications/:id/remint-invite` (PLATFORM_ADMIN + CSRF). Atomic in the approval tx: revoke the stale PENDING invite, mint a fresh one. Refuses a non-APPROVED application and an already-onboarded owner. Concurrent double-remint → retryable 409.
- **Web:** `remintInvite()` client + a **Regenerate** control on revealed rows that swaps the shown link in place.

## Slice C — operator-registration + review-card polish (`37640177` form + `143dc53b` card)
- **#5/#6/#11** Localized registration-form field errors, `aria-invalid`/`aria-describedby`/`role=alert`, fleet placeholder + `aria-required`.
- **#10** Fleet-size / business-type render **localized labels** via typed `Record<enum,key>` maps — never the raw storage codes (`6-20` / `INDIVIDUAL`). A new enum member is a compile error, not a leaked code.
- **#4** Approved invite links are accumulated **keyed by application id**, so approving a second row never drops the first row's still-uncopied link.

New i18n keys across all three locales (en/ja/zh); parity passes at 1459 keys. Deliberately skipped **#8** (duplicate-email enumeration oracle — intentional UX).

## Test plan
[...]

PR #1399, opened 2026-07-03
chore(#1371): cap admin applications list + distinguish slug-collision 409

jackli921:
Closes #1371

Two LOW housekeeping items from the #1350 review, one commit each, TDD.

## Slice A — distinguish the operators.slug clash from "already reviewed" (`cf9905a4`)

`approve()` mapped **every** `23505` inside the approval tx to `ConflictError('application already reviewed')`. A slug race between two *different* same-named businesses approved concurrently also lands there, so the loser saw a misleading "already reviewed" when the application is in fact still PENDING and a retry succeeds (the slug re-resolves).

- New `OPERATOR_SLUG_CONSTRAINT = 'operators_slug_unique'` constant (the migration gives the unique index a stable name).
- `approve()` matches that constraint by name and throws a distinct, retry-able 409; every other in-tx unique-violation keeps the generic message. Mirrors the two-part match already used in `submit()`.
- Tests: a slug-clash (constraint-named 23505) → `/retry/i`; a non-slug 23505 → `already reviewed`.

## Slice B — bound the admin list with keyset pagination (`1cf81528`)

`OperatorApplicationService.list` returned every matching row — an unbounded scan as the onboarding queue grows.

- Service now takes a required `limit` + optional `cursor` and returns `{ data, nextCursor }`, mirroring the `CustomerService` keyset pattern (overfetch-by-one, seek on `createdAt DESC, id ASC`).
- Route clamps `limit` via `parseLimit` (default 50, max 100) and keeps `data` a **bare array** with `nextCursor` in the envelope extras, so the web client's `z.array` parse is unchanged (backward-compatible).
- The opaque cursor is encoded/decoded in the **service** (not the repo) because operator-applications has two repo impls; each repo takes a structured `{ createdAt, id }` pivot — no cursor-string parsing duplicated across impls.
- Web admin client follows `nextCursor` to the end (via the existing `unwrapPage` helper) so the full pending queue still renders, with a page cap that refuses to truncate silently.

## Verification

- `bun run --filter @kuruma/api test` → **2616 passed**; `--filter @kuruma/web test` → **1899 passed**. Both packages typecheck; biome clean; pre-commit gate (biome + file-size + module-boundaries + tsc ×2) passed on each commit.
- New tests: slug vs generic 409 (service); keyset pagination across a page boundary with no overlap (service + route); web cursor-follow threading (`api.test.ts`).

## Residual risk

- The **drizzle** `list` keyset SQL is exercised only by typecheck + parity with the in-memory impl and the proven `CustomerRepository` pattern — there is no real-DB integration test for this specific query (the route tests run against in-memory repos). The cursor codec + overfetch logic (the parts most likely to be wrong) live in the service and are fully unit-tested.
- Cursor timestamp precision: `createdAt` is `defaultNow()` (microsecond) while the cursor carries millisecond ISO; a theoretical skip only occurs for two applications created in the same millisecond — not a real condition for an admin onboarding queue.

### DIFF
--- a/drizzle/0099_lonely_thor_girl.sql
+++ b/drizzle/0099_lonely_thor_girl.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "operator_applications" ALTER COLUMN "createdAt" SET DATA TYPE timestamp (3) with time zone;--> statement-breakpoint
+ALTER TABLE "operator_applications" ALTER COLUMN "createdAt" SET DEFAULT now();
--- a/drizzle/meta/0099_snapshot.json
+++ b/drizzle/meta/0099_snapshot.json
@@ -1,0 +1,6154 @@
+{
+  "id": "e3aae30f-7439-4afd-bd77-85537bb69029",
+  "prevId": "422de279-fd24-4976-b81f-dcc94a8ebff0",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.operator_applications": {
+      "name": "operator_applications",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "status": {
+          "name": "status",
+          "type": "operator_application_status",
+          "typeSchema": "public",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'PENDING'"
+        },
+        "businessName": {
+          "name": "businessName",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "contactName": {
+          "name": "contactName",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "contactEmail": {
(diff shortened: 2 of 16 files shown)


jackli921:
Heads-up: the develop-wide `e2e-real-db` failure (marketplace-happy-path step 6a) that was blocking this PR is fixed on develop as of #1403 (merged). It was a date-fragile test, not app code. Rebase/merge `origin/develop` to pick up the fixed spec and CI should go green.

jackli921:
Superseded by #1398 (merged to develop), which delivered both halves of #1371: the admin-queue bound (#1371a, limit+offset, capped <=100) and the slug-collision 409 distinction (#1371b, `operators_slug_unique` in pg-errors + approve() slug-race handling). Closing as redundant. The only unique parts here — keyset `{data,nextCursor}` pagination and a `timestamptz(3)` precision fix — serve a design the team didn't adopt (offset is sufficient for a <=100-row admin queue). No unique behavior is lost.

PR #1400, opened 2026-07-03
fix(#1260): bind operator-portal vehicle writes to the picked operator (slice 4)

jackli921:
## What

Slice 4 of #1260 (pre-GA authz: scope operator-portal writes to the picked operator). Binds every operator-portal **vehicle** write to the operator the caller is acting as, closing a bypass-admin hole where a platform admin could mutate any operator's vehicle by raw id.

## Why (the hole)

`FLEET_WRITE_ROLES` (BUSINESS_ROLES) admits PLATFORM_ADMIN and legacy STAFF/ADMIN.
The vehicle catalog is read under `operatorReadScope`, where all of those resolve `all` scope, so `repo.findById`/`findByIds` hands them ANY operator's vehicle by raw id.
The "read-only preview" that keeps a picker-admin scoped to one operator was enforced **client-side only** — a crafted request could update / retire / bulk-status / photo / maintenance-write another tenant's vehicle.

## Change

- **API guard** (`tenancy.ts` → `assertFleetWriteWithinOperator` / `fleetWriteDenialResult`): a caller that reached the target through an unscoped `all` read must name the operator it is acting as (`?operatorId=`), and that operator must own the target — else 422 `OPERATOR_REQUIRED` (no acting id) or 404 (wrong owner, no enumeration oracle). Row/tenant-scoped callers pass through untouched (read scope already clamped them). Keyed on `!isOperatorRole`, not `bypassScope`, because legacy STAFF is `all`-scoped-but-not-bypass here.
  - Applied to `VehicleService.update` / `softDelete` / `bulkUpdateStatus` / `toggleStatus`, vehicle-photo `uploadPhotos` / `deletePhoto`, and maintenance writes. Bulk is atomic — one cross-operator vehicle rejects the whole batch (no partial write).
- **Web vertical slice**: `withOperatorScope()` + `pickedOperatorId` threaded through all 6 fleet write calls and prop-drilled through the fleet components. Without this half, every picker-admin fleet edit would 422 — a **live regression** since #1264 made that path reachable. (Owner chose the full slice.)

## Tests

- New service-scope suites: `vehicle-write-scope`, `vehicle-photo-write-scope`, `maintenance-write-scope` — 422 / 404 / happy paths + legacy-STAFF mutation-resistance + atomic-bulk no-partial-write (asserts both the in-scope and out-of-scope vehicle stay untouched).
- Web `api.test.ts` + component tests assert `pickedOperatorId` reaches each write.

## Verification

- api: tsc, boundaries, biome pass; unit **2641** pass; integration **488** pass
- web: tsc, biome pass; tests **1908** pass

## Scope / follow-ups

- **Does not close #1260.** Vehicle writes is one path; the picker-admin manual-booking customer-scoping (slice 3) remains — `booking-creation.create` keys its customer-scope check on `isOperatorRole(ctx.role)`, which is false for PLATFORM_ADMIN, so a picker-admin can currently book for any renter.
- Follow-up (LOW): vehicle-photo routes gate `STAFF_ROLES` (= PLATFORM_ADMIN only), so an OPERATOR session 403s before reaching upload — the web-side "undefined for an operator session" comment is misleading, and it raises a product question (should operators upload photos at all?). Tracking separately.

Refs #1260


jackli921:
CI note: the red `e2e-real-db` check is the **known date-rollover flake #1401**, not a defect in this PR.

- Failure is `marketplace-happy-path.auth.spec.ts:140` step "6a. operator calendar shows the new booking" — a test that anchors the operator month view at a hard-coded `2026-07-15` and clicks `getByRole('button', { name: /Sarah Smith/ }).first()`. On 2026-07-03 the seed's `-5d` booking (`C3P9JKLM`) rolls into the July grid and `.first()` grabs it, so this run's booking code never shows.
- **Develop HEAD `8181c8ca` fails the identical test/line when re-run** (per #1401). This PR touches **zero** booking/manage/calendar files — only operator-fleet vehicle writes + the API guard. All other checks (`test-and-build`, `db-drift`, `e2e`) are green.
- Blocked on #1401's (test-only) fix landing on develop; will rebase + re-run `e2e-real-db` then.

PR #1415, opened 2026-07-03
feat(#1230): team picker (slice 6) — platform-admin manages a picked operator's team

jackli921:
## Summary

Closes the last gap in the operator-context picker (epic #1230, **slice 6**): a `PLATFORM_ADMIN` can now read and manage a **picked** operator's team (invite / revoke / deactivate) as that operator. `/operators/me/*` was self-scoped with no foreign-id surface; this threads an optional `?operatorId=` through the five routes and the web, honored **only** for the platform tier.

Implements the design + plan in #1412 (`docs/superpowers/specs|plans/2026-07-02-team-picker*`). **Refs #1230 — does not close the epic.**

## Architecture (one seam)

- Replace the service's private `requireOwnOperator(ctx)` with a pure **`resolveTeamOperatorId(ctx, inputOperatorId?)`** in `tenancy.ts` — an **allowlist**: operator role → own id (foreign input **ignored**, no cross-tenant); `PRIVILEGED_ROLES` (`PLATFORM_ADMIN`) → the input id (honored only here), no-pick → 422; everyone else (renter, PARTNER, legacy STAFF/ADMIN) → 403.
- Deliberately **stricter** than `resolveOperatorIdForWrite` (which keys on `!isOperatorRole` and would honor legacy STAFF/ADMIN) — pinned by a route test (G6a).
- No schema, no migration, no new authz gate.

## Slices / commits (TDD throughout)

**API**
- `resolveTeamOperatorId` resolver + reads thread `?operatorId=` (slice 1, pre-existing).
- `OperatorNotFoundError extends NotFoundError` → global handler maps a bogus picked id to **404, not 500** (c1). Base `NotFoundError.name` widened to `string` so the subclass can override (matching is by `instanceof`).
- Service writes accept a picked `operatorId` via the resolver; `requireOwnOperator` deleted. No-pick admin now **422** (was 403).
- Write routes thread `?operatorId=` + **real-pg integration test** (mint under picked X; cross-tenant deactivate → 404; last-owner → 409; audit names picked operator + admin actor).

**Web**
- `api.ts` threads `operatorId` into the 3 mutations + 2 reads; folds it into the query keys so a tenant switch refetches.
- The 3 dialogs take an `operatorId` prop.
- Team route registered in the picker; `business.team.pickOperatorPrompt` (en/ja/zh).
- `team.tsx`: capability-gated resolution + a **keyed `OperatorTeamData` child** — all-mode fires **no team read** (prompt only), and a tenant switch remounts (resets dialog/selection state). A retained `?operator=` is honored only for a picker-admin (a legacy STAFF/ADMIN's team read would 403).
- Ripple: `team` is now a picker route, so the "does not honor ?operator" fixtures in `operator-context.test.ts` / `BusinessLayout.test.tsx` move to the fleet by-id detail.

## Test plan

- [x] `bun run --filter @kuruma/api test` — 242 files / 2705 pass
- [x] `bun run --filter @kuruma/api test:integration` — 75 files / 497 pass (incl. new `operator-team-picker`)
- [x] `bun run --filter @kuruma/web test` — 285 files / 1955 pass
- [x] `bunx tsc --noEmit` clean in api / web / shared
- [x] `lint:boundaries` + `lint:modules` + `lint:size` exit 0; biome clean
- [x] i18n key parity (en/ja/zh)

## Security pins

- `resolveTeamOperatorId` is the deny point (not the owner-write gate): a legacy STAFF write-with-`?operatorId=` is **403** (G6a), proving team is not collapsed onto the looser write resolver.
- An `OPERATOR_STAFF` is still 403 even with a valid `?operatorId=` (owner-tier gate).
- CSRF pin (G6b): the query param cannot mask an absent/stale token — a 403 still throws.
- A bogus picked id is 404 (no existence oracle beyond the operator FK).

### DIFF
--- a/packages/api/src/auth/guards.ts
+++ b/packages/api/src/auth/guards.ts
@@ -53,7 +53,10 @@ export class ScopeRequiredError extends Error {
  * this doubles as the cross-tenant seal (no existence oracle).
  */
 export class NotFoundError extends Error {
-  readonly name = 'NotFoundError'
+  // Typed `string`, not the literal, so subclasses (e.g. OperatorNotFoundError,
+  // #1230 c1) can override with a distinct name; the global handler + local
+  // catches match by `instanceof`, never on this literal.
+  readonly name: string = 'NotFoundError'
   constructor(message = 'Not found') {
     super(message)
   }
--- a/packages/api/src/routes/operator-team.ts
+++ b/packages/api/src/routes/operator-team.ts
@@ -5,11 +5,13 @@ import type { OperatorTeamService } from '../services/operator-team'
 import { ok, parseBody } from './helpers'
 
 /**
- * #904: operator self-service team management. The tenant is ALWAYS the caller's
- * own operator — these are `/operators/me/*` paths with no id segment, so there
- * is no foreign-id surface to scope-check or leak. Owner-vs-member gating and the
- * absent-tenant seal live in `OperatorTeamService`; a denied write surfaces as a
- * 403 via the global `ForbiddenError` handler, so the routes stay HTTP-only.
+ * #904 / #1230: operator self-service team management. An operator always acts on
+ * its OWN tenant — a foreign `?operatorId=` is ignored. A PLATFORM_ADMIN using the
+ * operator-context picker names the target tenant via `?operatorId=` on the reads
+ * (`resolveTeamOperatorId` honors it only for that privileged tier). Owner-vs-member
+ * gating, the picker allowlist, and the absent-tenant seal live in
+ * `OperatorTeamService`; denials surface as 403/422 via the global handlers, so the
+ * routes stay HTTP-only.
  */
 export function createOperatorTeamRoutes(service: OperatorTeamService) {
   const app = new Hono()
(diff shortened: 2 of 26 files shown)


jackli921:
⚠️ **Keep epic #1230 open on merge.** This is slice 6 of the epic, not the whole epic. The branch contains a `fix(#1230): OperatorNotFoundError…` commit; on a **squash merge** GitHub scans the concatenated body for closing keywords and may auto-close #1230 (this bit #1260 on PR #1400). If it does, just `gh issue reopen 1230` — or merge-commit / trim that line from the squash body.

jackli921:
Heads-up from a duplicate implementation (PR #1419, now being closed in favor of this one — we independently executed the same slice-6 plan concurrently).

One small thing worth cross-pollinating: for the **c1 fix** (`OperatorNotFoundError` → 404), #1419 kept the change localized to `OperatorNotFoundError` and left `packages/api/src/auth/guards.ts` untouched. This PR widens the shared `NotFoundError.name` base type to `: string`. Both compile and behave identically (matching is `instanceof` everywhere, nothing reads `.name`), but the localized approach avoids changing the shared base class that every `NotFoundError` subclass inherits — slightly smaller blast radius if you want to adopt it. Non-blocking either way.

Also filed a standalone follow-up from the review: **#1420** (capability-gate the `?operator=` param in `settings.tsx` loader, mirroring what slice 6 did in `team.tsx`). Applies regardless of which PR merges.

jackli921:
✅ **code-reviewer pass: SHIP** — no CRITICAL/HIGH/MEDIUM findings.
- Authorization verified sound: allowlist resolver is genuinely stricter than `resolveOperatorIdForWrite`; owner-tier gate runs before the resolver on all 3 writes; bogus picked id → 404, no existence oracle.
- The `guards.ts` `NotFoundError.name` widening confirmed correct with **nil blast radius** (nothing branches on the `.name` literal; all matching is `instanceof`). LSP-clean vs a standalone class.
- Two LOW notes applied in `508f7e41`: `encodeURIComponent` on the `id` path segments; a comment documenting the intentional read-path `200 []` (not 404) so it isn't "fixed" into an existence oracle.

Also re-synced latest `develop` (clean, messages JSON auto-merged) — branch is current.

PR #1425, opened 2026-07-03
refactor(shared): fold payment_anomaly_* enums into enums.ts SSoT (#1383)

jackli921:
## What
Folds the `payment_anomaly_kind` and `payment_anomaly_resolution` value sets into the `enums.ts` single source of truth — the last domain enum that was still hand-copied.

## Why
The arrays were declared **twice**: inline literals fed to `pgEnum` in `db/payment.ts`, and again as `PAYMENT_ANOMALY_KINDS` / `PAYMENT_ANOMALY_RESOLUTIONS` in `types/payment-anomaly.ts`, linked only by a mirror test. Every other pgEnum already derives from `enums.ts`. Two literal copies = the drift class this SSoT exists to kill.

## How
- Move both arrays (+ derived types) into `enums.ts`, grouped with the other `payment_*` enums, docs carried over.
- Feed them to `pgEnum(...)` in `db/payment.ts` (identical values/order → **no SQL change**).
- Re-export from `types/payment-anomaly.ts` so the web boundary import path (`@kuruma/shared/types/payment-anomaly`) stays **stable** — web reads them as pure data since it cannot import the Drizzle schema.
- Pin both arrays with literals in `enums.test.ts` (the old mirror test is now tautological, so this literal pin is the real fence for the SSoT values).

## Out of scope
`packages/api/src/stores.ts` still declares a local `PaymentAnomalyKind` union (`'DOUBLE_PAYMENT' | 'AMOUNT_MISMATCH'`) — an api-internal copy, left untouched to keep this PR scoped to `shared` per the issue.

## Verification
- `bun run db:generate`: **"No schema changes, nothing to migrate"** (zero diff, no migration)
- `bun run db:verify`: 3 static drift checks pass; `drizzle/` untouched
- tsc: shared / api / web all clean (re-export contract holds)
- shared suite: **906 passed** (incl. new enum pins + the mirror test)

Part of #1369 (maintainability tracker, S3).

Closes #1383

### DIFF
--- a/packages/shared/src/db/payment.ts
+++ b/packages/shared/src/db/payment.ts
@@ -9,7 +9,12 @@ import {
   timestamp,
   uniqueIndex,
 } from 'drizzle-orm/pg-core'
-import { PAYMENT_EVENT_STATUSES, PAYMENT_REFUND_STATUSES } from '../enums'
+import {
+  PAYMENT_ANOMALY_KINDS,
+  PAYMENT_ANOMALY_RESOLUTIONS,
+  PAYMENT_EVENT_STATUSES,
+  PAYMENT_REFUND_STATUSES,
+} from '../enums'
 import { operators } from './auth'
 import { bookings } from './booking'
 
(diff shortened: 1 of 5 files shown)


jackli921:
Superseded by #1426 (`5a0916e9`), which landed the same payment_anomaly enum SSoT fold on develop first (concurrent session). Closing as redundant.

One bit here wasn't in #1426: the fold makes `tests/types/payment-anomaly.test.ts`'s two 'mirrors the DB enum' assertions tautological (both sides now derive from the enums.ts SSoT) and its 'only place the two are pinned together' comment false. Filing that as a small follow-up rather than racing a duplicate PR.

Issue #688, opened 2026-06-13
refactor(shared): derive enum types + Zod enums from pgEnum enumValues (SSoT foundation)


jackli921:
**Atomic unit (foundation)** · Scope: `packages/shared/**` only · Depends on: none · **Unblocks #696 (api) + #697 (web)** · Parallel-safe with all non-shared issues.

The single highest-leverage maintainability win (audit Theme 2, enum half — the role half landed via #683). The same closed sets are hand-copied with no compile-time link; `bookingStatus` alone is spelled out in 10+ places, and the schema test only checks the pgEnum so stale copies compile green and ship a runtime mismatch.

**This issue = the shared foundation only (so the two consumer PRs can then proceed independently):**
1. For each pgEnum, `export type X = (typeof xEnum.enumValues)[number]` (some already exist at `schema.ts:795`). Expose them through a **no-DB shared subpath** so web can import (web can't import api `stores.ts`).
2. Convert `shared/validators/*.ts` to `z.enum(<enum>.enumValues)` (currently hand-typed literals at `booking.ts:28,43`, `vehicle.ts:40,132,141`, `vehicle-class.ts:29`, `document.ts:29`).

Pattern already proven in-repo by `lib/luggage.ts` `LUGGAGE_SIZES`. Effort: S-M.

### Consumers (do after this lands, in parallel)
- [ ] #696 — api: convert `stores.ts` unions to derived
- [ ] #697 — web: replace bookingStatus literal copies with the shared import

---
_From 2026-06-13 maintainability audit (`docs/2026-06-13-maintainability-audit.md`)._

jackli921:
Merged via PR #755 (squashed to c33d8b1e on marketplace-pivot). Base≠default branch → closing manually and dropping in-progress.

PR #1426, opened 2026-07-03
refactor(#1383): fold payment_anomaly_* enums into enums.ts SSoT

jackli921:
## What

The `payment_anomaly_kind` / `payment_anomaly_resolution` value sets were spelled **twice** with no compile-time link:
- hand-inlined in `db/payment.ts`'s `pgEnum()` calls, and
- re-declared in `types/payment-anomaly.ts` (the web-boundary copy — web can't import the Drizzle schema).

This is exactly the hand-copied-literal drift class the `enums.ts` SSoT (#688) exists to kill. S3 of tracker #1369.

- **`enums.ts`** now owns `PAYMENT_ANOMALY_KINDS` / `PAYMENT_ANOMALY_RESOLUTIONS` (+ derived types), next to the other payment enums.
- **`db/payment.ts`** feeds those arrays to `pgEnum('payment_anomaly_kind', PAYMENT_ANOMALY_KINDS)` etc. (imported from `../enums`, mirroring the existing `PAYMENT_EVENT_STATUSES` wiring).
- **`types/payment-anomaly.ts`** re-exports the arrays + types from `../enums`, so web keeps importing them from `@kuruma/shared/types/payment-anomaly` unchanged. The wire-shape interfaces stay here.
- **`enums.test.ts`** gains the two `payment_anomaly_*` cases in the pgEnum↔SSoT order-pin table (#688). The existing `payment-anomaly.test.ts` mirror now holds *structurally* — both sides derive from one array, so drift is impossible.

## Why it's safe

Value order is preserved exactly (positional pgEnum). No schema change:

```
$ bun run db:generate
No schema changes, nothing to migrate 😴   # drizzle/ clean, zero diff
```

Consumers' types are unchanged (`PaymentAnomalyKind` etc. resolve to the same unions).

## Verification

- `db:generate` → **zero diff** (no `CREATE TYPE` change).
- shared suite **906 pass**; `shared` / `api` / `web` tsc clean.
- Pre-commit (biome, lint:size, lint:boundaries, tsc×2) green.

Closes #1383. Part of #1369.


jackli921:
Closing as a duplicate of #1425 (both fold `payment_anomaly_*` into the `enums.ts` SSoT and close #1383). Two parallel sessions took the same issue.

Picked **#1425** as the winner on test strategy: it adds a **literal value pin** in `src/enums.test.ts` (`expect(PAYMENT_ANOMALY_KINDS).toEqual(['DOUBLE_PAYMENT','AMOUNT_MISMATCH'])`) — a mutation-resistant fence that catches any value/order change — and correctly **removes the now-tautological mirror test** in `tests/types/payment-anomaly.test.ts` (once both sides derive from one array, `expect(X).toEqual(X)` proves nothing). This PR's additions (the mirror test + the #688 pgEnum↔array row) are both tautological after the fold, so the unit layer would have no real fence on the values.

Nice touch worth carrying over to #1425 (optional): adding the `payment_anomaly_*` rows to the #688 `pgEnum.enumValues === array` table in `tests/enums.test.ts` for registry consistency with the other folded enums.

Reopen if #1425 stalls — the work here is equivalent and salvageable.

jackli921:
Correction to my comment above (written mid-race): **this PR (#1426) is the one that actually merged** (`5a0916e9`) and resolved #1383 — a parallel session landed it while I was reviewing, and #1425 got closed. My earlier note had the winner backwards.

One thing worth a small follow-up: #1425's test approach was stronger — a **literal value pin** in `src/enums.test.ts` (`expect(PAYMENT_ANOMALY_KINDS).toEqual(['DOUBLE_PAYMENT','AMOUNT_MISMATCH'])`) plus removal of the now-**tautological** mirror test in `tests/types/payment-anomaly.test.ts`. What landed here keeps both tautological checks (they compare the array to a pgEnum built *from* that array), so the unit layer has no real fence on the values — only db-drift CI would catch a bad edit. Worth porting that literal pin.

PR #1421, opened 2026-07-03
feat(#464): renter class-combo booking flow (book a class, operator assigns the car)

jackli921:
## Summary

Closes the last gap in #464: a renter can now **discover a vehicle-CLASS deal on a store page and book it** (no specific car chosen); the operator assigns a concrete car on/before pickup (already built). Entry point is the owner-locked **Option A** — a class deal links to the store page, which lists the store's class deals alongside its cars.

Implements `docs/plans/2026-07-02-464-renter-combo-booking-plan.md` (2 architect passes) end to end, one vertical slice per commit.

### Server (no migration, no new endpoint)
- Extract the CLASS_COMBO producer into `ClassOfferingService` (out of `FlatSearchService`), inject it into **both** flat-search and `StorefrontDetailService`; wire in the composition root.
- `GET /storefronts/:locationId/vehicles` now returns `classOfferings`, scoped to the store's `operatorId` + location (cross-operator isolation test pins the scoping).

### Web
- Store-page **"Class deals"** section (`ClassOfferingCard`), schema mirror pinned with `satisfies z.ZodType<ClassComboSearchResult>` + `.default([])` for deploy skew.
- Reservation loader resolves a vehicle-or-class `ReservationSubject`; the wizard books either, priced off the class rate plan, and shows "a specific car will be assigned before pickup".
- Booking client sends a `CLASS_COMBO` discriminant via a **distributive** `CreateBookingDraft` (avoids the non-distributive `Omit`-over-union trap).
- Confirmation shows the class + note pre-assign, and the concrete car post-assign via an expanded `?expand=vehicle,renter` read (variant query; base reads untouched).

## Test plan
- [x] `ClassOfferingService` + storefront-detail unit tests (happy path, ACRISS narrowing, sold-out exclusion, cross-operator isolation — mutation-verified)
- [x] Web unit: class card, loader subject resolution, wizard CLASS_COMBO submit (asserts `classId` present / `requestedVehicleId` absent), null-hourly pricing floor, sold-out-at-submit surfaces, confirmation pre/post-assign (guard mutation-verified)
- [x] Full API unit (2729) + **integration (495, incl. storefront-detail contract change)**
- [x] Web unit (2007), shared, typecheck ×2, boundaries, module lint, i18n parity (1496 keys)
[...]

PR #1431, opened 2026-07-03
fix(#1430): forbid one-way CLASS_COMBO bookings (dropoff must equal pickup)

jackli921:
## Summary

Closes #1430. A defensive follow-up from the code review of PR #1421's read-side turnaround fix (`cbd603f3`).

**The gap.** The class-combo availability card and booking creation size their occupancy window from **different turnaround sources**:
- **Read-side** (`ClassOfferingService.toCombo`) sizes `[from, effectiveEndAt)` by the **pickup** storefront's turnaround — the only one it can see at browse time (no dropoff chosen yet).
- **Write-side** (`booking-creation.ts` `resolveDropoffEffectiveEnd`) sizes it by the **dropoff** location's turnaround, and the combo path fully supports `dropoff != pickup`.

So a one-way combo could be advertised as available and then fail at checkout (or the inverse). Not UI-reachable — the renter wizard and manual-booking dialog both hardcode `dropoff == pickup` — but **the API is source-agnostic by design (Trip.com hits the same routes)**, so a 3rd-party `CLASS_COMBO` body with a different-turnaround dropoff triggers it. The validator previously accepted it on purpose.

**The fix.** Forbid one-way class-combo bookings at the validator boundary: a `.refine()` on `createBookingSchema` requires `dropoffLocationId === pickupLocationId` when `fulfillmentMode === 'CLASS_COMBO'` (path `dropoffLocationId` → 400). SPECIFIC bookings are unaffected (they carry a concrete vehicle, not a class float). The read model can only reason about the pickup turnaround, so this codifies what the system can actually honor.

**Reversible:** if one-way combos are wanted later, the read model must change too (size conservatively by dropoff turnaround); this constraint would be lifted then.

## Test plan

- [x] TDD: RED (combo with `dropoff != pickup` → asserts reject + error path `dropoffLocationId`) → GREEN (refine).
- [x] Combo validator fixtures pinned to `dropoff == pickup` via a new `comboCommon`; SPECIFIC keeps `common` (one-way still allowed there).
- [x] `@kuruma/shared` full suite 907, `@kuruma/api` booking + combo consumers 353 (route combo POSTs use same-location, unaffected), typecheck shared/api/web, biome, file-size, module-boundaries.
- No migration. No web change (web already sends `dropoff == pickup` for combos).

## Out of scope

- SPECIFIC one-way bookings (per-vehicle availability; separate concern if a parallel divergence exists).

### DIFF
--- a/packages/shared/src/validators/booking.ts
+++ b/packages/shared/src/validators/booking.ts
@@ -82,6 +82,22 @@ export const createBookingSchema = z
     message: 'Provide either renterId or walkInCustomer, not both',
     path: ['walkInCustomer'],
   })
+  // #1430: a CLASS_COMBO must be returned to its pickup location. The renter
+  // books off an availability card that can only size the occupancy window by
+  // the PICKUP turnaround (no dropoff is chosen at browse time), while booking
+  // creation sizes it by the DROPOFF turnaround (booking-creation.ts
+  // resolveDropoffEffectiveEnd). Allowing a one-way combo lets the card and the
+  // write-side disagree ("advertised but sold out at checkout"). One-way combos
+  // are unsupported product-wide, so the contract forbids them here. SPECIFIC
+  // bookings are unaffected (they carry a concrete vehicle, not a class float).
+  .refine(
+    (data) =>
+      data.fulfillmentMode !== 'CLASS_COMBO' || data.dropoffLocationId === data.pickupLocationId,
+    {
+      message: 'Class-combo bookings must be returned to the pickup location',
+      path: ['dropoffLocationId'],
+    },
+  )
 
 export const updateBookingStatusSchema = z.object({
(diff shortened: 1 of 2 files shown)


jackli921:
Independent code-reviewer pass: **clean, no CRITICAL/HIGH/MEDIUM.** Verified the boundary is complete (POST /bookings via `createBookingSchema` is the only client-supplied combo-create path — no bypass), the refine is correctly scoped to CLASS_COMBO (SPECIFIC unaffected), `dropoffLocationId` is required so an omitted dropoff can't slip through, and the tests are mutation-resistant (assert rejection + error path). No shared-`CreateBookingInput` type ripple; web unaffected. One LOW noted and intentionally left: the predicate is a case-sensitive UUID compare — academic (Postgres normalizes uuid case, both fields come from the same client, and it fails in the safe direction).

Issue #1437, opened 2026-07-03
feat(catalog i18n): operator-authored custom items + shared-catalog kill-switch


jackli921:
## Summary

Deliver owner requirements 3 and 4 from the operator-catalog design: **operators can author their own add-on/insurance items with independent en/ja/zh translations**, and **platform admins can turn the shared catalog OFF platform-wide** without affecting anything operators created themselves.

Design (owner-reviewed, v2.3): `docs/plans/2026-07-03-operator-custom-items-and-catalog-killswitch.md`

## Owner decisions (final review, 2026-07-03)

- **A1** — "off" = hide/freeze/reversible (templates persist; picked items keep rendering).
- **D1** — Insurance independent translations are **IN SCOPE now** (slice 3), not deferred.
- **D3** — When off, the admin template library is **fully hidden**.
- **D5** — Operators **can edit** a self-authored item's translations after create.

## Slices (each a vertical, shippable PR)

1. **Add-on self-authored path** — `nameI18n` column + CHECK, validator (exactly-one refine), resolver/service/dedupe branches, persistence plumbing, operator "custom item" form, audit type split (§8.1). Ships FIRST (creation must have a non-catalog path before the switch can hide the picker).
2. **`SHARED_CATALOG` kill-switch** — registry `serverOnly`/`serverDefault` (danger zone, §6.2), server seam, picker endpoint + create reject `templateId` when off, web hides picker + admin library. Depends on slice 1.
3. **Insurance independent translations** — `nameI18n` + resolver + wire-contract promotion + operator form (§8.3). Own multi-PR track.

## Design principle

Additive, not a rewrite: a row is EITHER *picked* (`templateId` set, renders from shared template — admin translations still propagate) OR *self-authored* (`nameI18n` set, operator owns the bundle). The shared catalog stays the translation-quality layer while ON.

PR #1459, opened 2026-07-04
feat(#1437): SHARED_CATALOG kill-switch (slice 2 of 3)

jackli921:
Slice 2 of 3 for #1437 (operator-authored catalog items + shared-catalog kill-switch). Follows slice 1 (operator self-authored add-ons, #1445). Design: \`docs/plans/2026-07-03-operator-custom-items-and-catalog-killswitch.md\`.

## What

Adds a platform kill-switch, \`SHARED_CATALOG\`, that turns off the shared add-on template catalog. When off, operators must self-author add-ons (the escape hatch from slice 1); the picker and platform template library disappear.

## How

- **Registry (`@kuruma/shared`).** New `SHARED_CATALOG` flag typed as a `serverOnly` entry (`serverDefault: true`) — it has no `VITE_FEATURE_*` build-time env; the API owns it and the web floors to `serverDefault`. Registry retyped `Record<Key, FeatureFlagEntry>` so indexed `.env` access stays uniform.
- **API enforcement.** `FeatureFlagsService.isEnabled(key)` is the single server-side seam (admin override wins, else registry `serverDefault`). `AddOnTemplateService.listForPicker` returns `[]` when off; `AddOnService.create` returns **422** (well-formed, authorized, catalog just disabled) if a `templateId` is picked while off. Both services take a `() => Promise<boolean>` thunk defaulting to `true` (fail-OPEN) so existing callers/tests are untouched; the composition root injects the real flag-backed thunk.
- **Web surface.** `resolveFeatureFlag` floors `serverOnly` keys to `serverDefault` on every fail-safe path. `AdminSidebar` hides the template-library link; `AddOnForm` hides the picker + identity toggle and forces the self-author path in CREATE when off (EDIT unaffected).

## Verification

- `tsc --noEmit` ×3 (shared/api/web) clean
- API import boundaries + web module boundaries clean
- Tests: shared 9/9, api 37/37 (incl. an end-to-end wired-route test proving the composed app empties the picker at 200 and rejects a template pick at 422), web 26/26 — **72 passing**

## Scope

`Refs #1437` (non-closing — epic stays open for slice 3: insurance purely self-authored).

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -486,8 +486,17 @@ export function createApp(overrides?: AppOverrides, repos: Repos = buildRepos(ov
   const vehicleService = new VehicleService(vehicleRepo, resolveWriteOperatorId, photosPublicUrl)
   const locationService = new LocationService(locationRepo, bookingRepo, cachedGeocoder, regionRepo)
   const insuranceOptionService = new InsuranceOptionService(insuranceOptionRepo)
-  const addOnService = new AddOnService(addOnRepo, addOnTemplateRepo)
-  const addOnTemplateService = new AddOnTemplateService(addOnTemplateRepo, addOnRepo)
+  const featureFlagsService = new FeatureFlagsService(featureFlagRepo)
+  // #1437: SHARED_CATALOG is server-enforced. Both the operator picker read (empty when
+  // off) and the add-on create path (reject a templateId when off) gate on ONE narrow
+  // thunk (ISP) rather than the whole FeatureFlagsService.
+  const isSharedCatalogEnabled = () => featureFlagsService.isEnabled('SHARED_CATALOG')
+  const addOnService = new AddOnService(addOnRepo, addOnTemplateRepo, isSharedCatalogEnabled)
+  const addOnTemplateService = new AddOnTemplateService(
+    addOnTemplateRepo,
+    addOnRepo,
+    isSharedCatalogEnabled,
+  )
   const templateLibraryService = new TemplateLibraryService(
     addOnTemplateRepo,
     insuranceTemplateRepo,
@@ -524,7 +533,6 @@ export function createApp(overrides?: AppOverrides, repos: Repos = buildRepos(ov
   )
   const reviewAggregateService = new ReviewAggregateService(reviewRepo)
   const reviewListService = new ReviewListService(reviewRepo)
-  const featureFlagsService = new FeatureFlagsService(featureFlagRepo)
 
   // Chain .route() calls so TypeScript infers the full route type tree.
   // hc<AppType> needs this to produce typed client methods.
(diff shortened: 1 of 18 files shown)


jackli921 (commented):
## Review — slice 2 `SHARED_CATALOG` kill-switch (reviewed against the architect plan for #1437)

Strong implementation. The load-bearing pieces are all correct:

- **Polarity trap closed.** `resolveFeatureFlag` floors a `serverOnly` key to `serverDefault`, never the envless build-time reader; `feature-flags-runtime.test.ts` locks the critical empty-map → ON case *and* proves flooring is scoped to `serverOnly` (CANCELLATION still falls to its `false` reader).
- **Server seam single-sources the default** (`isEnabled = overrides[key] ?? serverDefault`), so the web flooring and server enforcement can't drift.
- **Composition root** reordered correctly; one narrow ISP thunk injected into both services. `shared-catalog-killswitch-wired.test.ts` boots the real app and proves *both* the picker-empty and create-reject paths observe the switch — the exact regression the unit tests can't catch (the gate is an optional ctor arg defaulting to ON, so dropping it in `index.ts` compiles clean and passes every unit test). That test is the highlight of this PR.
- **Registry danger zone** updated by *filtering* `serverOnly`, not loosening assertions; Finding-A (TS2339 on generic `.env` access) solved cleanly via the `Record<Key, FeatureFlagEntry>` widening.

### 🔴 Requesting one change — admin template library is deep-linkable when the catalog is off

`packages/web/src/routes/$locale/_admin/admin/templates.tsx` gates nothing at the loader. `AdminSidebar` drops the nav link, but the route still runs `ensureQueryData(templateLibraryQueryOptions())` and renders the full library on a direct visit to `/{locale}/admin/templates`. Owner requirement **D3** is "admin template library *fully hidden*" when the catalog is off, so nav-drop alone doesn't satisfy it — a platform admin (or a stale bookmark) still reaches it.

Fix (mirrors the existing `_admin.tsx` `beforeLoad` redirect precedent):

```ts
beforeLoad: async ({ context, params }) => {
  const map = await context.queryClient.ensureQueryData(featureFlagsQueryOptions())
  if (!resolveFeatureFlag(map, 'SHARED_CATALOG'))
    throw redirect({ to: '/$locale/admin', params: { locale: params.locale } })
},
```

Both `featureFlagsQueryOptions` and `resolveFeatureFlag` are exported from `@/vite/config`. Worth a small route-redirect test: a `{ SHARED_CATALOG: false }` override redirects; an absent/`true` map renders.

### Optional (non-blocking)

- **The 422 create-reject carries no `code`.** 403 → 422 was the right move (it's a state, not an authz denial). Adding `code: 'SHARED_CATALOG_DISABLED'` would let the web localize the reject copy via `failResult`'s `code` passthrough instead of matching the English string — low value while the picker is hidden, but nearly free.
- **`FeatureFlagEntry` is flat-optional**, so `serverOnly ⇒ serverDefault` is guaranteed by the runtime "every serverOnly has a boolean `serverDefault`" test rather than by the type. That's a fine fence — just noting it's the test, not the compiler; a discriminated union would move it to compile time.
- **A regression test asserting `updateAddOnSchema` has no `templateId` field** would guard against a future PATCH widening silently reopening picked-while-off.

PR #1510, opened 2026-07-08
test(#1489): pin cold-load redirect chain does not steal announcer focus

jackli921:
## What

A #1489 follow-up. The #1489 review raised a theoretical WCAG 3.2.1 concern: a `beforeLoad` `throw redirect()` on **cold load** might be misread as a post-mount navigation and steal focus to the `RouteAnnouncer`'s invisible anchor.

## Finding — not a bug on the installed router

Reproduced against the installed `@tanstack/react-router` (1.170.15) with a real memory-history router. Every app redirect lives in `beforeLoad`, so an entire redirect chain is caught **mid-load** as one continuous pending phase (verified in `Transitioner.js`: `isAnyPending` stays true across every hop). `resolvedLocation` therefore settles **exactly once**, on the final target, and the hook already seeds that single settle as the initial load. **No focus steal, no runtime fix needed.**

## What this PR adds

A real-router regression **pin** (chained `/` -> `/mid` -> `/dest`) asserting:
1. the chain settles exactly once (`onResolved` fires once, on the final target) — the diagnostic, and
2. focus is left on the document root, not yanked to the announcer — the user-facing guard.

A future router upgrade that changes the settle-count (re-opening the focus-steal) fails loudly. Test-only; no runtime change.

## Verification

- Sabotage-verified: removing the hook's initial-load seed turns the pin red.
- `RouteAnnouncer.redirect.test.tsx` green; full web suite 2165/2165; web+api `tsc`, biome, lint:modules/size all clean.
- Rebased onto latest `develop` (picks up #1509's `RouteAnnouncer`-wraps-`Outlet` change; the pin mirrors the real wrapping mount).

Refs #1489. Not a `Closes` — #1489 is already closed; this is a follow-up.

## Test plan

- [x] `bunx vitest run src/vite/nav/RouteAnnouncer.redirect.test.tsx`
- [x] Full web suite
- [x] Sabotage check (seed removed -> red)


jackli921:
Abandoning this follow-up per owner request. Item 2 was a regression-pin test only (the predicted focus-steal did not reproduce on the installed router). Recoverable from commit 621754e2 if ever revived.

Issue #1205, opened 2026-06-27
Messaging: build operator inbox + multi-tenant participants before un-gating (#1032 follow-up)


jackli921:
## Why

In-app messaging shipped **renter-side only** under #1032 (threads + messages schema, renter inbox/thread/composer, auto-thread-on-booking via `ensureThread`). The booking-scoped, Turo-style model is correct. But two pieces were never built, so the feature is **gated off in beta** behind `VITE_FEATURE_MESSAGING` and only previewable by the platform admin (per #1161).

This issue tracks the work required **before the flag can be un-gated** for operators.

Design doc: `docs/plans/2026-04-14-messaging-design.md` (de-staled in #1203).

## Scope — un-gate blockers

1. **Operator inbox (the headline gap).** Operators have no messaging UI today — they would only learn of a renter message by email. Build `/manage/messages` (inbox with unread badges) + a `business-nav-items` entry + the thread view/composer on the operator side. The renter-side components exist and can be mirrored.

2. **Multi-tenant participants.** Replace the single global recipient — `process.env.DEFAULT_STAFF_ID` lookup in `packages/api/src/index.ts` (the `staffUserId = process.env.DEFAULT_STAFF_ID` line, ~356; doc cites the older `:342`) — with **operator-derived participants**: the thread counterpart is the booking's operator (`booking.operatorId`), and **any staff of that operator** (`OPERATOR_OWNER` / `OPERATOR_STAFF`) can read and reply. Operator-as-*org*, not a single hardcoded user. The `DEFAULT_STAFF_ID` env var is a single-tenant artifact that breaks in the marketplace.

   > This "operator-as-org, not per-user" call is the **same decision reviews faces** (#1158 — operator-side review deduped per operator, not per staff-user). Settle it once for both features.

3. **Operator first-unread email.** Email the operator's notifiable address(es) on first unread message in a thread (was "no email — owner is in the app daily", which no longer holds with N operators and no operator inbox).

## Out of scope

Postgres RLS (the design doc records the application-level-scoping decision and the #1119/#1124 audits that hardened it; RLS remains a later belt-and-suspenders option).

Refs #1032 (renter-side ship), #1161 (beta gating), #1158 (shared operator-as-org decision with reviews).

Issue #328, opened 2026-04-17
security(api): scope findByIdempotencyKey to caller (cross-tenant leak risk)


jackli921:
**From architect audit (main @ ef1dd8c).**

`packages/api/src/repositories/drizzle/thread.ts:121` and `message.ts:65` — `findByIdempotencyKey(key)` takes no `ctx`. Callers in `routes/messages.ts:69,93` feed a client-supplied key straight in and return the record on match.

## Attack
An authenticated renter who learns (or brute-forces a weak) idempotency key reads threads/messages they do not participate in.

## Learn: Missing Scope on Idempotency Lookup
Idempotency-key reads have no ownership filter on the row.
Cross-tenant data leak the moment keys collide, are logged/leaked, or use low entropy.
**Heuristic:** every lookup that returns a tenant-owned row must be scoped — idempotency keys are inputs, not authz.

## Fix
Add `ctx` and filter by participant (threads) / sender (messages); OR enforce `(senderId, key)` uniqueness and filter by `ctx.userId`.

jackli921:
starting

PR #1522, opened 2026-07-11
fix(messaging): GA hardening tail from the tenant-isolation review

jackli921:
Implements the 6 non-blocking hardening findings from the messaging tenant-isolation security review, filing off the rough edges before the `MESSAGING` feature flag flips on for GA. The isolation model itself was already reviewed SOUND (no CRITICAL/HIGH). **This PR does not flip the flag.**

Refs #1476 (GA epic — stays open).

## The 6 fixes

1. **Drop `idempotencyKey` from the message read model** (`1438cd51`). It was echoed to every thread participant on `GET /threads/:id` and gave a 500-vs-201 existence oracle for another sender's key. Removed from the `Message` type, `messageColumns`, `RawMessageRow`, `normaliseMessage`, the DISTINCT-ON SELECT, and the in-memory shaping. The write path (`sendMessageSchema`, `create()` param, the DB column + unique index, the in-memory dedup index) is intentionally unchanged.

2 & 4. **Per-user rate limits on translate + send** (`0c87b3d6`). The global `RATE_LIMITER` is per-IP, which NAT collapses. Added `MESSAGE_TRANSLATE_LIMITER` (20/60s) on `POST /messages/:id/translate` (each uncached (message, language) hits an external provider) and `MESSAGE_SEND_LIMITER` (30/60s) on `POST /threads/:id/messages` (thread flooding). Optional `AppOverrides` seams; resolved in `index.ts`; new `wrangler.toml` bindings (namespace_id 1007/1008). Mirrors `routes/vehicle-photos.ts`. A composition-root test drives `createApp` end-to-end and guards against a send/translate limiter swap.

3. **Cross-sender idempotency-key collision → 409** (`0a214f4d`). The key is globally unique but `findByIdempotencyKey` is sender-scoped (#328), so a different sender replaying another sender's key escaped as a raw `UNIQUE_VIOLATION` 500. Now a clean `ConflictError` → 409.

5. **Push thread-list limit/offset into the query** (`510c9251`). `listThreads` loaded every in-scope thread via `findAll()` then sliced in memory — a `PLATFORM_ADMIN` `all` scope meant every thread on the platform per `GET /threads`. New `ThreadRepository.findPage(ctx, {limit, offset})` pushes `LIMIT/OFFSET` + `COUNT(*)` into SQL (Drizzle) with an in-memory mirror, ordered newest `createdAt` first. `findAll()` is untouched (user-directory needs the full set); its hydration is extracted into a shared helper.

6. **Refresh a stale user-directory comment** (`cb5a790f`). Operator messaging ships now (#1205); the operator self-only directory scoping is a deliberate isolation choice, not a not-built-yet placeholder. Comment-only.

`36fc49ab` folds in two LOW precision fixes from the code-review pass (collation-independent in-memory tiebreaker; honest 409 comment).

## Verification

- tsc clean ×3 (shared, api, web).
- `lint:boundaries` OK; `lint:modules` soft-warn only (pre-existing web-tree baseline, untouched here).
- api unit **2970/2970**; full integration **520/520** real-pg; web **2208/2208**.
- Every fix is TDD (RED → GREEN); the two guard tests (limiter swap, findPage offset pushdown) are sabotage-verified.
- Fresh `code-reviewer` pass: **SHIP** — no CRITICAL/HIGH/MEDIUM; both LOW findings addressed in `36fc49ab`.

## Not in scope

- Does not flip the `MESSAGING` flag.
- No schema migration (none of the 6 fixes touch the schema).

### DIFF
--- a/packages/api/src/app-overrides.ts
+++ b/packages/api/src/app-overrides.ts
@@ -109,6 +109,12 @@ export type AppOverrides = {
   photoUploadUserLimiter?: RateLimitBinding
   publicCatalogLimiter?: RateLimitBinding
   operatorApplicationLimiter?: RateLimitBinding
+  // Per-user caps on the authenticated messaging surface (messaging GA, #1476):
+  // send throttles POST /threads/:id/messages; translate throttles the external
+  // translation provider on POST /messages/:id/translate. Absent ⇒ the
+  // globalThis-resolved CF binding, or unthrottled local dev.
+  messageSendLimiter?: RateLimitBinding
+  messageTranslateLimiter?: RateLimitBinding
   // Over-limit ⇒ the geocoder skips the lookup (#574). Inject a deny-binding in
   // tests; absent ⇒ the globalThis-resolved GEOCODE_LIMITER (or unthrottled dev).
   geocodeLimiter?: RateLimitBinding
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -210,6 +210,14 @@ export function createApp(overrides?: AppOverrides, repos: Repos = buildRepos(ov
     ((globalThis as Record<string, unknown>).OPERATOR_APPLICATION_LIMITER as
       | RateLimitBinding
       | undefined)
+  const messageSendLimiter =
+    overrides?.messageSendLimiter ??
+    ((globalThis as Record<string, unknown>).MESSAGE_SEND_LIMITER as RateLimitBinding | undefined)
+  const messageTranslateLimiter =
+    overrides?.messageTranslateLimiter ??
+    ((globalThis as Record<string, unknown>).MESSAGE_TRANSLATE_LIMITER as
+      | RateLimitBinding
+      | undefined)
 
   const translationProvider = createTranslationProvider()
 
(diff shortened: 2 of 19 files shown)


jackli921:
## Architect review — verdict: Ship (no CRITICAL/HIGH)

Ran an independent architect-review pass (structural anti-patterns) on top of the earlier code-reviewer SHIP. Explicitly cleared the two highest-risk items:

- **Middleware ordering — VERIFIED SAFE.** The limiter keyFunc `(c) => requireUser(c).id` provably runs *after* `requireAuth` (registered at `index.ts` before the sub-app mounts; Hono composes in registration order, same structure vehicle-photos ships in prod). Even in the counterfactual, `requireUser` throws → 500, never an auth bypass.
- **The participant-join `as` cast — JUSTIFIED.** `.select(threadColumns)` fixes the flat projection regardless of the join; `toThread` re-validates at the boundary. Identical to the sanctioned `findByIdempotencyKey` idiom.

Other findings all LOW/INFO (non-transactional COUNT+page is inherent to offset pagination and harmless at scale; read-model `idempotencyKey` drop breaks no consumer — web zod parser never read it; boundaries/DI clean).

Two notes for the record:
- **Translate limiter budget is per-*request*, not per-provider-call** — it runs before the handler so it also counts `cached:true` hits. Over-throttle only (fail-safe); size the 20/60s as requests/user.
- **Follow-up #1524** filed (LOW, deferred, not a GA blocker): `findPage` sorts by `createdAt DESC` but `threads.createdAt` is unindexed — add `idx_threads_createdAt_id`. Migration-only; deliberately kept out of this PR to avoid a 0107-class migration collision with the other in-flight PRs.

